PR #4, opened 2018-02-19
EN-801 language choice

mcarta-rt:
Created language selection integration on designtime-admin-app project.


src/ui/login/LoginFormContainer.js
```diff
@@ -15,6 +16,7 @@ export const mapStateToProps = state => ({
 // map the props
 export const mapDispatchToProps = dispatch => ({
```
alessioscalici: just FYI (I'm not requiring this change): this could be rewritten as:
```javascript
export const mapDispatchToProps = {
    performLogin,
    setLanguage: setCurrentLanguage,
};
```

PR #26, opened 2018-03-06
moved sass outside of src, and module resolver in webpack

alessioscalici:


.eslintrc.js
```diff
@@ -20,7 +20,7 @@ module.exports = {
     "settings": {
         "import/resolver": {
             "node": {
-                "moduleDirectory": ["node_modules", "src"]
+                "moduleDirectory": ["node_modules", "src", "sass"]
```
nicpuddu: why is the sass directory being checked by eslinter?

PR #27, opened 2018-03-06
En 877

alessioscalici:
- sass-lint fixed
- removed watch-css and build-css scripts
- uninstalled node-sass-chokidar


nicpuddu:
:shipit:

PR #32, opened 2018-03-07
En 775 list fragments with paginator

nicpuddu:


src/state/pagination/reducer.js
```diff
@@ -0,0 +1,64 @@
+import { SET_PAGE } from 'state/pagination/types';
+
+const initialState = {
+  page: 1,
+  pageSize: 10,
+  lastPage: 1,
+};
+
+const isAttributeANumber = attribute => (
```
fedelocci: move in a file inside util directory ?

nicpuddu: done!

PR #35, opened 2018-03-08
En 879 page configuration grid

alessioscalici:
- Added a basic page and route for page configuration
- Updated page model state to have a selected page model
- Added a selector returning a rows/columns structure from the selected page model in the state
- Added a component that renders the grid based on the structure from the selector


src/api/pageModels.js
```diff
@@ -1,10 +1,18 @@
...
   })
 );
+
+export const getPageModel = code => (
+  // call GET /pagemodels/<code>
+  new Promise((resolve) => {
+    setTimeout(() => {
+      resolve(PAGE_MODEL, code);
+    }, 500);
+  })
```
nicpuddu: why didn't you use throttle?

src/state/page-models/selectors.js
```diff
@@ -1,3 +1,132 @@
...
+      }
+    }
+  });
+  return matrix;
+};
+
+const printMatrix = (matrix) => {
+  const yMax = matrix.length ? matrix[0].length : 0;
+  let row = [];
+  const rowstr = [];
```
nicpuddu: constants before lets

src/state/page-models/selectors.js
```diff
@@ -1,3 +1,132 @@
...
+    }
+  }
+  return {
+    x1: xStart, x2, y1, y2, frame: matrix[xStart][y1],
+  };
+};
+
+const getCols = (matrix, bounds) => {
+  let xCur = bounds.x1;
+  const cols = [];
```
nicpuddu: const before lets

src/state/page-models/selectors.js
```diff
@@ -1,3 +1,132 @@
...
+};
+
+const getPageModelGridStruct = (pageModel) => {
+  if (!pageModel) {
+    return null;
+  }
+  const yMax = pageModel.configuration.frames
+    .reduce((acc, frame) => (acc < frame.sketch.y2 ? frame.sketch.y2 : acc), 0);
+  const matrix = createFramesMatrix(pageModel.configuration.frames);
+  printMatrix(matrix);
```
nicpuddu: lets not initialize a constant that is not reused

src/state/page-models/selectors.js
```diff
@@ -1,3 +1,132 @@
...
+  const bounds = {
+    x1: 0, x2: 11, y1: 0, y2: yMax,
+  };
+
+  const stack = [bounds];
+  while (stack.length) {
+    const curBounds = stack.pop();
+    const rows = getRowsWithChildren(matrix, curBounds);
+    const isLeaf = curBounds !== bounds && rows.length === 1 && rows[0].cols.length === 1;
+    if (!isLeaf) {
```
nicpuddu: isLeaf is only used in the if, we may get rid of the constant

src/ui/pages/config/PageConfigGrid.js
```diff
@@ -0,0 +1,82 @@
...
+    return (
+      <div
+        key={`row-${row.y1}`}
+        className="PageConfigGrid__row"
+      >
+        {renderedCols}
+      </div>
+    );
+  });
+};
```
nicpuddu: lets split this into more methods it is really hard to read it in its current state

PR #42, opened 2018-03-14
En 887 integrate login api

nicpuddu:


entando-service:
All is well

src/api/apiManager.js
```diff
@@ -18,31 +18,81 @@ const defaultPage = { page: 1, pageSize: 10 };
...
 export const config = (reduxStore) => {
   store = reduxStore;
 };
 
 export const makeMockRequest = (request, page = defaultPage) => {
   validateRequest(request);
-  return new Promise(resolve => throttle(() => resolve(buildResponse(request.mockResponse, page))));
+  const errors = getErrors(request.errors);
+  return new Promise(resolve => throttle(() => (
+    resolve(new Response(
```
alessioscalici: Nice way to have a realistic mock response :)

src/api/login.js
```diff
@@ -1,11 +1,25 @@
...
-    reject(BODY_ERROR);
-  }
+export const getErrors = (username, password) => (
+  (username === 'admin' && password === 'adminadmin') ?
+    [] :
+    [{ code: 1, message: 'invalid credentials' }]
+);
+
+const login = (username, password) => makeRequest({
+  uri: '/entando-sample/OAuth2/access_token',
```
alessioscalici: Probably `/entando-sample/` should not be placed here. Should the right place be the `DOMAIN` variable in `.env` file?

src/state/login-form/actions.js
```diff
@@ -21,10 +20,16 @@ const ERROR_LOGIN_MESSAGE = 'error: username or password is invalid';
 export const performLogin = (username, password) => (dispatch) => {
   if (username && password) {
     dispatch(setLoginErrorMessage(''));
-    login(username, password).then(() => {
-      gotoRoute('dashboard');
+    return login(username, password).then((response) => {
+      if (response.ok) {
+        response.json().then(() => {
+          gotoRoute('dashboard');
```
alessioscalici: I know it was already like this, but could you use the constant `ROUTE_DASHBOARD` from `app-init/router` here?

entando-service:
All is well

src/state/login-form/actions.js
nicpuddu: sure thing!

src/api/login.js
nicpuddu: I'm torn on this one:
the attribute is named `DOMAIN` and that is not part of it.
On one hand the repetition would be endless tho.

I can make the change if you want to, though

alessioscalici: The problem is we'll have many server instances, and their URI will unlikely begin with `entando-sample`, so I think it's a good idea to have a neutral API path here, and set domain and path in the .env file.

You are right that's not part of the domain, so we could either rename DOMAIN in something like API_URI, or add another variable for the path. What do you think?

entando-service:
All is well

PR #52, opened 2018-03-20
En 930 edit user screen

fedelocci:


entando-service:
All is well

src/ui/users/common/UserForm.js
```diff
@@ -13,6 +13,25 @@ import FormLabel from 'ui/common/form/FormLabel';
...
+export const renderStaticField = (field) => {
+  const { input, label, name } = field;
+  const fieldValue = (input.value.title) ? input.value.title : input.value;
+  if (!input.value) {
+    return null;
+  }
+
+  return (
+    <div className="form-group">
+      <label htmlFor={name} className="control-label col-sm-2">
```
alessioscalici: col-sm-2 -> col-xs-2

src/ui/common/form/SwitchRenderer.js
```diff
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import { Switch } from 'patternfly-react';
 
 const SwitchRenderer = ({ input }) => {
-  const switchValue = typeof input.value === 'string' ? (input.value === 'true') : input.value;
+  const switchValue = typeof input.value === 'string' ? (input.value === 'true' || input.value === 'active') : input.value;
```
alessioscalici: At the moment, SwitchRenderer only handles boolean values. If you want to extend the selection to other types, you can do something like:
```
const SwitchRenderer = ({ input, trueValue, falseValue }) => {
  const switchValue = input.value === 'true' || input.value === true || input.value === trueValue;
  return (<Switch
    {...input}
    value={switchValue}
    onChange={(el, val) => input.onChange(val ? trueValue : falseValue)}
  />);
};


SwitchRenderer.propTypes = {
  /* eslint-disable react/forbid-prop-types */
  trueValue: PropTypes.any,
  falseValue: PropTypes.any,
  /* eslint-enable react/forbid-prop-types */
  input: PropTypes.shape({
    value: PropTypes.oneOfType([PropTypes.string, PropTypes.bool]).isRequired,
    onChange: PropTypes.func.isRequired,
  }).isRequired,
};

SwitchRenderer.defaultProps = {
  trueValue: true,
  falseValue: false,
};
```

and when using it:

```
<Field
  component={SwitchRenderer}
  trueValue="active"
  falseValue="inactive"
```

entando-service:
All is well

src/api/user.js
```diff
@@ -0,0 +1,20 @@
...
+import { makeRequest, METHODS } from 'api/apiManager';
+
+export const getUser = (params) => {
+  if (params) {
+    console.info(`calling API /user/${params}`);
+  }
+  return makeRequest({
+    uri: '/users/',
+    method: METHODS.GET,
+    mockResponse: USER_OK.payload,
```
alessioscalici: Please refactor USER_OK removing the payload layer, because it's already handled by makeRequest

src/api/user.js
```diff
@@ -0,0 +1,20 @@
...
+  });
+};
+
+export const putUser = user => makeRequest({
+  uri: '/users/',
+  method: METHODS.PUT,
+  mockResponse: USER_OK,
```
alessioscalici: Here is ok, but the generated response will be `{ payload: { payload: { ... user data ...} } }`.

After refactoring USER_OK mock removing the payload layer, this will be correct

src/state/users/actions.js
```diff
@@ -16,3 +22,35 @@ export const fetchUsers = (page = 1, params) => dispatch =>
     dispatch(setUsers(data.payload));
     dispatch(setPage(data.metaData));
   });
+
+
+export const fetchUserForm = username => dispatch =>
+  getUser(username).then((response) => {
+    if (response.ok) {
+      response.json().then((json) => {
```
alessioscalici: You would need to return the response.json() promise in order to keep the promise chained:
`return response.json().then((json) => {`

BUT since we adopted the convention to return a `new Promise()` in each new API and explicitly resolving it, you'd better use that for this API too (as you did in `sendPutUser`)

entando-service:
All is well

PR #53, opened 2018-03-20
EN 968 groups page

mcarta-rt:


entando-service:
All is well

src/api/groups.js
```diff
@@ -1,8 +1,21 @@
...
+import { GROUPS_OK_PAGE_1, GROUPS_OK_PAGE_2 } from 'test/mocks/groups';
+import throttle from 'util/throttle';
 
-// eslint-disable-next-line
-export const getApiGroups = () => (
-  new Promise((resolve) => {
-    resolve(GROUPS);
-  })
-);
+export const getGroups = (page, params) => new Promise((resolve) => {
```
alessioscalici: Needs to be aligned with the new way we manage API and API tests (see https://github.com/entando/designtime-admin-app/pull/54)

entando-service:
All is well

entando-service:
All is well

entando-service:
All is well

src/state/groups/actions.js
```diff
@@ -1,16 +1,26 @@
...
   },
 });
 
 // thunk
-export const fetchGroups = () => dispatch =>
-  getApiGroups().then((data) => {
-    dispatch(addGroups(data.payload));
+export const fetchGroups = (page = { page: 1, pageSize: 10 }, params = '') => dispatch =>
+  new Promise((resolve) => {
+    getGroups(page, params).then((response) => {
```
alessioscalici: What if the promise get rejected? The Promise would remain pending forever.
You need to add `.catch()` blocks in internal promises, something like

```javascript
export const fetchGroups = (page = { page: 1, pageSize: 10 }, params = '') => dispatch =>
  new Promise((resolve, reject) => {
    getGroups(page, params).then((response) => {
      if (response.ok) {
          dispatch(setGroups(data.payload));
          resolve();
        })
        .catch(reject);
      } else {
        resolve();
      }
    })
    .catch(reject);
  });
```

PR #60, opened 2018-03-23
En 991 page configuration header

alessioscalici:


entando-service:
All is well

test/state/page-config/actions.test.js
```diff
@@ -204,7 +230,116 @@ describe('state/page-config/actions', () => {
...
+  describe('setSelectedPageOnTheFly()', () => {
+    beforeEach(() => {
+      store = mockStore(INITIAL_STATE);
+    });
+
+    it('dispatches REMOVE_PAGE_WIDGET if value = false', (done) => {
+      getSelectedPageModelMainFrame.mockReturnValue({ pos: 3 });
+      store.dispatch(setSelectedPageOnTheFly(false)).then(() => {
+        const actionTypes = store.getActions().map(action => action.type);
+        expect(actionTypes).toEqual([REMOVE_PAGE_WIDGET]);
```
nicpuddu: I understand why you are doing this, but I don't know if computationally is any better than just checking the array manually

test/state/page-config/actions.test.js
```diff
@@ -204,7 +230,116 @@ describe('state/page-config/actions', () => {
...
+        expect(actionTypes).toEqual([SET_PAGE_CONFIG]);
+        done();
+      }).catch(done.fail);
+    });
+
+    it('does nothing if there is no pageCode parameter', (done) => {
+      getParams.mockReturnValue({});
+      store.dispatch(restoreSelectedPageConfig()).then(() => {
+        expect(restorePageConfig).not.toHaveBeenCalled();
+        expect(store.getActions().length).toBe(0);
```
nicpuddu: you can just use `.toHaveLength` in this kind of tests

test/state/page-models/selectors.test.js
```diff
@@ -78,4 +78,66 @@ describe('state/page-models/selectors', () => {
...
+  describe('getSelectedPageModelDefaultConfig', () => {
+    it('returns null if there is no selected page model', () => {
+      const state = buildStateWithSelectedPageModel(null);
+      expect(getSelectedPageModelDefaultConfig(state)).toBe(null);
+    });
+
+    it('returns a page config array containing null for frames with no default widget', () => {
+      const state = buildStateWithSelectedPageModel(SINGLE_CELL_PAYLOAD);
+      expect(getSelectedPageModelDefaultConfig(state)).toEqual(expect.arrayContaining([null]));
+    });
```
nicpuddu: shouldn't this check that the array **only** contains `null`?

test/state/page-config/actions.test.js
alessioscalici: Not computationally, but it's a way to easily read the sequence of actions being dispatched.
Do you think it would be better to check manually?

entando-service:
All is well

test/state/page-config/actions.test.js
nicpuddu: it's ok either way really :stuck_out_tongue:

nicpuddu:
:shipit:

PR #67, opened 2018-03-26
EN-1081 updated LoginForm container with prop currentLanguage

mcarta-rt:


entando-service:
All is well

entando-service:
All is well

PR #68, opened 2018-03-27
En 964 languages and labels

alessioscalici:


entando-service:
All is well

entando-service:
All is well

entando-service:
All is well

src/state/labels/actions.js
```diff
@@ -64,3 +71,18 @@ export const createLabel = label => dispatch => (
...
+        response.json().then(() => {
+          dispatch(removeLabelSync(label));
+          resolve();
+        });
+      } else {
+        resolve();
+      }
+    });
+  })
+);
```
fedelocci: i this case if thare are error is not calling.
i should be like this?

export const removeLabel = label => dispatch => (
  new Promise((resolve) => {
    deleteLabel(label).then((response) => {
      
        response.json().then(() => {
if (response.ok) {
          dispatch(removeLabelSync(label));
          resolve();
} else {
dispatch(addErrors(json.errors.map(err => err.message)));
        resolve();
      }
        });
      
    });
  })
)

entando-service:
All is well

entando-service:
All is well

entando-service:
All is well

PR #77, opened 2018-03-29
add link from  kebabMenu list to detail group

fedelocci:
add link from  kebabMenu list to detail group


entando-service:
All is well

PR #106, opened 2018-04-13
add API, test , reducer and asction for delete fragment

fedelocci:
Fragment delete not making an API call


entando-service:
All is well

PR #108, opened 2018-04-16
En 1131 user role detail

alessioscalici:


mcarta-rt:
retest this please

alessioscalici:
retest this please

entando-service:
retest this please

entando-service:
ok to test

entando-service:
retest this please

src/state/roles/selectors.js
```diff
@@ -1,10 +1,25 @@
...
+export const getUserRefs = state => state.roles.selected.userReferences;
 
 export const getRolesList = createSelector(
   [getRolesMap, getRolesIdList],
   (rolesMap, idListRoles) => idListRoles.map(id => (rolesMap[id])),
 );
+
+export const getSelectedRolePermissionsList = createSelector(
+  [getSelectedRolePermissions, getPermissionsList],
+  (rolePermissions, permissions) => permissions.filter(permission =>
```
nicpuddu: there's no need to use square brackets for this

src/state/roles/selectors.js
```diff
@@ -1,10 +1,25 @@
...
+    rolePermissions[permission.code]).map(permission => permission.descr),
+);
+
+export const getSelectedRole = createSelector(
+  [getSelectedRolePermissionsList, getRoles],
+  (permissions, roles) => ({ ...roles.selected, permissions }),
```
nicpuddu: same applies here

src/ui/common/references/UserRefsTable.js
```diff
@@ -0,0 +1,99 @@
...
+  userReferences: PropTypes.arrayOf(PropTypes.shape({
+    username: PropTypes.string,
+    fullName: PropTypes.string,
+    lastLogin: PropTypes.string,
+    status: PropTypes.string,
+  })),
+};
+
+UserRefsTable.defaultProps = {
+  onWillMount: null,
```
nicpuddu: in the onWillMount method of the component it does not check whether or not onWillMount is defined before calling it as a function, although the default is null.
If the parameter is being passed by the container anyway it should just be required and have no default, or if the check does not happen it should default to an empty function

PR #111, opened 2018-04-16
implemented fronted libraries

nicpuddu:


__mocks__/@entando/utils.js
```diff
@@ -0,0 +1,22 @@
+const utils = jest.genMockFromModule('@entando/utils');
+const real = require.requireActual('@entando/utils');
+
+utils.formattedText = jest.fn().mockImplementation(id => id);
+utils.isInteger = real.isInteger;
```
alessioscalici: You could implement these functions as mocks and still keep their implementation like this:

`utils.isInteger = jest.fn(real.isInteger);`

This way you keep the real behaviour, but you can still use `expect(func).toHaveBeenCalled()` against these functions

nicpuddu: alternatively I could just add the spy itself to the method.
I'm fine about making the change but if you don't mind I'd do ti on a separate PR.

PR #137, opened 2018-04-26
En 1300 integrations

nicpuddu:
the API itself does not exist that, that is why the fetch is not implemented.
It will be done in a separate PR before closing the actual 1300 ticket.


entando-service:
All is well

src/state/rootReducer.js
```diff
@@ -57,6 +58,7 @@ const reducerDef = {
   router,
   users,
   widgets,
+  integrations,
```
fedelocci: how i can see the reducers are ordered :smile:

PR #149, opened 2018-05-03
En 1394 right hand side overlaps

alessioscalici:
- Fixed widget list overlapping the config grid on small screens
- Fixed case sensitive widget search
- Fixed duplicated key issue (category code / widget code used as key were colliding)


entando-service:
All is well

entando-service:
All is well

PR #153, opened 2018-05-07
update version package @entando/utils and @entando/apimanager

fedelocci:
update version libraries


entando-service:
All is well

PR #164, opened 2018-05-10
EN-1413 fix page add button

mcarta-rt:


entando-service:
All is well

src/ui/pages/common/RoleListTable.js
```diff
@@ -1,118 +0,0 @@
...
-};
-
-RoleListTable.defaultProps = {
-  onWillMount: null,
-  loading: false,
-  roles: [],
-  onClickDelete: null,
-};
-
-export default RoleListTable;
```
nicpuddu: you deleted the component but I don't see any import to this component being deleted, was this a mistake?

nicpuddu: it was a duplicated file

PR #166, opened 2018-05-10
EN-1415 fix search pages

mcarta-rt:
If search returns no results, searchPages will be void array [].
If clearSearch is dispatched, searchPages will be null.
The difference between void array and null is useful switching between PageTree and PageListSearchTable.


entando-service:
All is well

PR #167, opened 2018-05-11
En-1384-file browser

alepintus:
Hi @nicpuddu  @alessioscalici @fedelocci ,
File browser page added,
Could you please take a look?

Thanks!


entando-service:
All is well

entando-service:
All is well

entando-service:
All is well

src/ui/file-browser/list/FileBreadcrumb.js
```diff
@@ -0,0 +1,117 @@
...
+        <Col xs={12}>
+          <Breadcrumb>
+            {renderRootFolderBreadcrumb()}
+            {renderProtectedFolderBreadcrumb()}
+            {renderItemsLinksBreadcrumb()}
+            {renderCurrentItemBreadcrumb()}
+          </Breadcrumb>
+        </Col>
+      </div>
+    );
```
nicpuddu: this should be inside the render and each method should be reworked.
You are splitting every single time from the say array instead of traversing it properly.

If we wanna merge like this it will have to go through major refactoring.

src/ui/file-browser/list/FilesListTable.js
```diff
@@ -0,0 +1,190 @@
...
+    );
+  }
+
+  render() {
+    return (
+      <div className="FilesListTable">
+        <Spinner loading={!!this.props.loading}>
+          {this.renderTable()}
+        </Spinner>
+      </div>
```
nicpuddu: same thing applies here

some of these could actually be their own components

src/ui/file-browser/list/FilesListTable.js
```diff
@@ -0,0 +1,190 @@
...
+          return 'false';
+        case 'protected':
+          return 'true';
+        default:
+          return this.props.pathInfo.protectedFolder;
+      }
+    };
+
+    const getQueryString = (path) => {
+      switch (path) {
```
alessioscalici: This code is equivalent, and I think it could be more readable:
```javascript
if (['protected', 'public'].includes(path)) {
  return null:
}
return path;
```

alepintus:
Hi @alessioscalici @nicpuddu ,
code changed as suggested and  new tests added.
Could you please have a look?

Thanks!

entando-service:
All is well

PR #172, opened 2018-05-15
EN-1418 fix user status icon, fix display user edit info correctly, f…

mcarta-rt:
…ix hide action view profile if user has no profile


entando-service:
All is well

nicpuddu:
this PR has code from 171. Some corrections were requested on that PR.

PR #177, opened 2018-05-16
Allow plugin forms them to modify the config object before submit

alessioscalici:
To be merged **before** the following PR: https://github.com/entando/entando-components/pull/239

Force plugin forms to wrap the forms with reduxForm themselves, in order to allow them to modify the config object before submit


entando-service:

entando-service:
All is well

PR #180, opened 2018-05-17
EN-1483 refactor group reference management

mcarta-rt:


entando-service:
All is well

src/state/groups/selectors.js
```diff
@@ -2,16 +2,77 @@ import { createSelector } from 'reselect';
 import { getLocale } from 'state/locale/selectors';
 
 export const getGroups = state => state.groups;
-
 export const getGroupsIdList = state => state.groups.list;
-
 export const getGroupsMap = state => state.groups.map;
+export const getGroupsTotal = state => state.groups.total;
```
fedelocci: use createSelector

fedelocci: use createSelector because it's memoized

src/api/groups.js
```diff
@@ -24,8 +20,8 @@ const getGroupErrors = groupname => (
     [{ code: 1, message: 'invalid group name' }]
 );
 
-const getErrorsReferences = (ref, groupname) =>
-  (ref[groupname] ? [] : [{ code: 1, message: 'invalid group name' }]);
+// const getErrorsReferences = (ref, groupname) =>
```
alessioscalici: Let's not leave comments around, can you please remove it?

src/state/groups/actions.js
```diff
@@ -194,56 +152,57 @@ export const sendDeleteGroup = groupCode => dispatch => (
...
-export const fetchCurrentReferenceContents =
-        fetchCurrentReference(getContentReferences, setSelectedGroupContentReferences);
-
-export const fetchCurrentReferenceResources =
-        fetchCurrentReference(getResourceReferences, setSelectedGroupResourceReferences);
+export const fetchCurrentPageGroupDetail = () => (dispatch, getState) => (
+  new Promise((resolve) => {
+    const { groupname } = getParams(getState());
+    getGroup(groupname).then((response) => {
+      response.json().then((json) => {
```
alessioscalici: You might have a non-json response, please make sure it is before calling response.json(), or use a try/catch block to handle these kind of errors

src/state/groups/selectors.js
alessioscalici: Actually, in this case it's not needed, it even causes more bad than good.
The selector as is, makes a simple access to an object that will always be populated, so it's computationally very cheap, and there is no calculation.
If you use createSelector, you'll create a memoized selector with a single result cache, meaning that you have some disadvantages:
1. you will call one function more each time you access the selector: you have to call the selector function, plus the wrapper function that checks the arguments and either returns the memoized or recalculates the result
2. you will create a cache, that's basically a constantly synchronized copy of state.groups.total, and a waste of memory
3. the code will be more verbose and less readable.

In these cases, we should use memoized selectors only when there is an access to a state from another domain, in that case using createSelector makes sense because in case of refactoring, it would be less error prone

entando-service:
All is well

entando-service:
All is well

PR #201, opened 2018-05-22
page models toasts

nicpuddu:
fixed also redirect bug when the update or creation would error

### DIFF
--- a/src/state/page-models/actions.js
+++ b/src/state/page-models/actions.js
@@ -9,7 +9,7 @@ import {
 import { setPage } from 'state/pagination/actions';
 import { addErrors } from 'state/errors/actions';
 import { toggleLoading } from 'state/loading/actions';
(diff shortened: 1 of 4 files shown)


entando-service:
All is well

src/state/page-models/actions.js
```diff
@@ -142,45 +151,41 @@ export const initPageModelForm = () => (dispatch, getState) => {
...
-          });
-        } else {
-          resolve();
-        }
+export const createPageModel = pageModel => dispatch => new Promise((resolve) => {
+  postPageModel(pageModel).then((response) => {
+    if (!response.ok) {
+      response.json().then((data) => {
+        dispatch(addErrors(data.errors.map(err => err.message)));
+        resolve();
```
fedelocci: you can remove this. You can call resolve after else

nicpuddu: no I cannot, this one resolves after another promise, while the else is not waiting for that one.

PR #208, opened 2018-05-23
EN-1504 updated user selector, updated authority table

mcarta-rt:


entando-service:
All is well

src/state/users/selectors.js
```diff
@@ -21,8 +21,8 @@ const getGroupRolesComboValue = state => formValueSelector('autorityForm')(state
 const getArrayData = array => (groups, roles) =>
   array.reduce((acc, item) => {
     acc.push({
-      group: { code: groups[item.group].code, name: groups[item.group].name },
-      role: { code: roles[item.role].code, name: roles[item.role].name },
+      group: item.group ? { code: groups[item.group].code, name: groups[item.group].name } : {},
```
alessioscalici: I have 2 questions:
1. What's inside `item.group` and `item.role`? Because if it's the code, you can use those values instead of `groups[item.group].code`
2. Why are we using `reduce` to return an array from an array? I think `map` would be much more appropriate

entando-service:
All is well

PR #210, opened 2018-05-23
EN-1455 route to user profile edit

mcarta-rt:
User list menu action updated, route to user profile edit updated.
IMPORTANT: user profile edit page has not been completed yet.


entando-service:
All is well

PR #220, opened 2018-05-25
EN-1510 Fix user restrictions

mcarta-rt:
success and error messages updated, userSettings state managment updated, fixed AttributeListMenuActionsProfile bug on shallow component


entando-service:
All is well

PR #236, opened 2018-05-29
EN-1463 fix datatype edit attribute code

mcarta-rt:
fix dataType edit attribute code and ProfileType/DataType status icon


entando-service:
All is well

PR #238, opened 2018-05-29
EN-1461-reference-status-icon

fedelocci:
create api, component, action and test


entando-service:
All is well

entando-service:
All is well

PR #257, opened 2018-05-31
fix title and amend selectors for errors during load the page

fedelocci:


entando-service:
All is well

entando-service:
All is well

PR #276, opened 2018-06-05
[5.0.1] EN-1584-501-profiles-types-edit-attribute-code-not-editable

alepintus:
Code readonly for profile type attributes in edit mode and removed comment in PageForm


entando-service:
All is well

PR #299, opened 2018-06-07
[5.0.1]- EN-1587-pageTree-active-languages

fedelocci:
add fields active languages


entando-service:
All is well

alessioscalici:
As the twin PR https://github.com/entando/app-builder/pull/300, please remove the language codes from the translation files

src/ui/common/form/FormLabel.js
```diff
@@ -18,13 +18,20 @@ const FormLabel = ({
...
+  ) : null;
+
   const fieldHelp = helpId ? (
     <FieldLevelHelp content={formattedText(helpId)} />
   ) : null;
 
   return (
     <span className="FormLabel">
       {langLabel}
+      {langLabelWithText}
```
alessioscalici: This could be better understandable and less error (label duplication) prone if you write:

``` {langLabel || langLabelWithText} ```

This way, you will always show one and only one labels, even if you provide 2 (in this case, precedence is given to the translated label)

entando-service:
All is well

fedelocci:
@alessioscalici  i've fixed the suggestion

PR #300, opened 2018-06-07
[MASTER] EN-1587-pageTree-active-language

fedelocci:
add fields active languages


entando-service:
All is well

src/locales/it.js
```diff
@@ -18,6 +18,281 @@ export default {
     'app.unpublish': 'Depubblica',
     'app.en': 'en',
     'app.it': 'it',
+    'app.aa': 'aa',
```
alessioscalici: All these codes don't need a translation. Why are you including them in the translation files? Wouldn't a constant be enough?

src/locales/en.js
```diff
@@ -18,6 +18,281 @@ export default {
     'app.unpublish': 'Unpublish',
     'app.en': 'en',
     'app.it': 'it',
+    'app.aa': 'aa',
```
alessioscalici: All these codes don't need a translation. Why are you including them in the translation files? Wouldn't a constant be enough?

entando-service:
All is well

entando-service:
All is well

PR #307, opened 2018-06-08
[MASTER] EN-1598-disable-not-usable-actions

fedelocci:
amend behavior menu for unpublish and delete


entando-service:
All is well

PR #317, opened 2018-06-11
[5.0.1] En 1573 add page code field validation 

fedelocci:
En 1573 add page code field validation 5.0.1


entando-service:
All is well

PR #322, opened 2018-06-12
[MASTER] En 1598 disable not usable actions master

fedelocci:


entando-service:
All is well

src/state/pages/selectors.js
```diff
@@ -92,14 +92,23 @@ export const getPageTreePages = createSelector(
...
-        ...pages[pageCode],
-        ...PAGE_STATUS_DEFAULTS,
-        ...pagesStatus[pageCode],
-        depth: getDepth(pages, pageCode),
-        isEmpty: !(pageChildren[pageCode] && pageChildren[pageCode].length),
-        title: pagesTitles[pageCode][locale],
-      }))),
+      .map((pageCode) => {
+        const isEmpty = !(pageChildren[pageCode] && pageChildren[pageCode].length);
+        let pagePublish = false;
```
alessioscalici: I think "pagePublish" does not reflect the nature of this property (indicating if the page has children which are published)
Could you rename it as "hasPublishedChildren" ?

src/ui/pages/common/PageTreeActionMenu.js
```diff
@@ -24,10 +24,17 @@ class PageTreeActionMenu extends Component {
...
+      disabled = true;
+    }
+    if (page.expanded) {
+      disabled = page.pagePublish;
+    }
     const changePublishStatus = page.status === PAGE_STATUS_PUBLISHED ?
       (
         <MenuItem
-          disabled={page.status === PAGE_STATUS_PUBLISHED && !page.isEmpty}
+          disabled={disabled}
```
alessioscalici: This is not the change that was required in the ticket IMHO.

> Disable "unpublish" when the page has even only one public child page
> Disable "delete" whatever its status, when a page has children pages

So, I think you should : 

1. in the UNPUBLISH button: ` disabled={page.hasPublishedChildren} `
1. in the DELETE button: ` disabled={!page.isEmpty} `

entando-service:
All is well

PR #346, opened 2018-06-18
[5.0.1] EN-1598-disable not usable action

fedelocci:
fix behavior publish action

### DIFF
--- a/src/state/pages/selectors.js
+++ b/src/state/pages/selectors.js
@@ -4,6 +4,7 @@ import { getLocale } from 'state/locale/selectors';
 import { PAGE_STATUS_PUBLISHED } from 'state/pages/const';
 import { getDomain } from '@entando/apimanager';
 import { PREVIEW_NAMESPACE } from 'ui/pages/config/const';
+import { get } from 'lodash';
 
 export const getPages = state => state.pages;
 export const getPagesMap = state => state.pages.map;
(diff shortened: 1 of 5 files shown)


entando-service:
All is well

src/state/pages/selectors.js
```diff
@@ -95,6 +95,7 @@ export const getPageTreePages = createSelector(
       .map((pageCode) => {
         const isEmpty = !(pageChildren[pageCode] && pageChildren[pageCode].length);
         let hasPublishedChildren = false;
+        const parentStatus = pages[pageCode].parentCode && pages[pages[pageCode].parentCode].status;
```
alessioscalici: Are you sure `pages[pageCode]` and `pages[pages[pageCode].parentCode]` can't be undefined here? Wouldn't it be better to use a more forgiving approach, like using lodash `get` and checking if the result is defined?

src/ui/pages/list/PageTreePageContainer.js
```diff
@@ -13,6 +13,7 @@ export const mapStateToProps = state => ({
 
 export const mapDispatchToProps = dispatch => ({
   onWillMount: () => {
+    dispatch(clearErrors());
```
alessioscalici: This will not show server errors from other pages (like Add page).
If you add a page with an already used code, you won't see the error. This problem has been recently fixed

PR #401, opened 2018-07-03
Embedded domain v5.0.0

kspokas:
Changes to allow AppBuilder to run as a sub-domain, and to connect to a relative path for the Mapp engine using the new EMBEDDED_DOMAIN environment variable


entando-service:
All is well

src/app-init/router.js
```diff
@@ -81,79 +81,79 @@ routerConfig(
...
-      { name: ROUTE_PLUGIN_CONFIG_PAGE, path: '/plugin/:pluginId/config' },
-      { name: ROUTE_DATABASE_LIST, path: '/database' },
-      { name: ROUTE_DATABASE_ADD, path: '/database/add' },
-      { name: ROUTE_DATABASE_REPORT, path: '/database/report/:dumpCode' },
-      { name: ROUTE_DATABASE_DUMP_TABLE, path: '/database/report/dumpTable/:dumpCode' },
-      { name: ROUTE_FILE_BROWSER, path: '/file-browser' },
-      { name: ROUTE_FILE_BROWSER_UPLOAD, path: '/file-browser/upload' },
-      { name: ROUTE_FILE_BROWSER_CREATE_FOLDER, path: '/file-browser/create-folder' },
-      { name: ROUTE_FILE_BROWSER_CREATE_TEXT_FILE, path: '/file-browser/create-text-file' },
+      { name: ROUTE_HOME, path: `${process.env.PUBLIC_URL}/` },
```
alessioscalici: I think this might be less invasive and error prone, if we just added a new config option to this router config object, e.g.:

```pathPrefix: process.env.PUBLIC_URL,```

and modified the router to add it to the paths of the compiled routes RegExps

Comment (kspokas) [reply src/app-init/router.js]: I should have left more context to what's going on here:

The goal of these changes is to allow us to package an "all-in-one" distribution of Entando that has a pre-built App Builder bundled with the Java application.  To do this, we need to add a fix to the App Builder to enable it to run in sub-folder and to use a relative path for the DOMAIN.  I wanted to make this not break the current way of doing things either.

1. To run in a sub-folder, I am adding the PUBLIC_URL to the route table.  PUBLIC_URL is a special env variable that will automatically be set up.  We only need to provide a value if we want it to run in a subfolder (example below).

2. To allow a relative path to the Mapp engine, I introduced a new env variable called EMBEDDED_DOMAIN, which can refer to the context of the Mapp engine.  If this is set, it will dynamically append the value of this variable to the base url of the application.  If not, the DOMAIN variable will work as before.

Here are the env files for example:

Example 1 ("as before"):
USE_MOCKS=false
DOMAIN=http://localhost:8081/entando

Example 2 (Embedded example):
USE_MOCKS=false
PUBLIC_URL=/appbuilder
EMBEDDED_DOMAIN=entando

Example 2 would result in the following working (wherever it is deployed):
http://server.name:port/appbuilder
http://server.name:port/entando

Hope this clears things up

Comment (kspokas) [reply src/app-init/router.js]: Here is a sample application that uses this method when using mvn jetty:run (and also includes a custom landing page at the root url): https://github.com/kspokas/entando-pam-sample/tree/appbuilder-experiment

Comment (alessioscalici) [reply src/app-init/router.js]: Makes sense to me. I'd like @nicpuddu to review as well before merging

File: package.json [hunk 2]
```diff
@@ -34,7 +34,7 @@
     "extract-text-webpack-plugin": "3.0.2",
     "file-loader": "1.1.5",
     "flux-standard-action": "^2.0.1",
-    "frontend-common-components": "github:entando/frontend-common-components#master",
+    "frontend-common-components": "github:entando/frontend-common-components#v5.0.0",
```
nicpuddu: it should be pointing to version 5.0.1

src/app-init/router.js
nicpuddu: I do agree with @alessioscalici on this, I'd rather not have the process variable used over and over, it is more error prone.

in theory it would be better adding a prefix parmeter to the router component in general, so that we don't have to deal with this in each route, and set it when configuring the router.

nicpuddu:
we also need to create the PR for master (v5.1.0) I'm also slightly confused because this PR is meant for 5.0.0 (according to the title) but it is patching 5.0.1

package.json
kspokas: I will omit changes to package.json from the PR, they are not relevant to what is in the update

src/app-init/router.js
kspokas: Makes sense, I will refactor this and update the readme

PR #402, opened 2018-07-03
[master] EN-1674 Limiting widget form titles to backoffice languages

alessioscalici:
- Replaced system languages fields with backoffice languages
- Unit tests


entando-service:
All is well

PR #415, opened 2018-07-05
[MASTER] EN-1504 Manager authorizaton users

fedelocci:
amend the behavior in the contaier and thunk when delete an authority


entando-service:
All is well

PR #441, opened 2018-07-19
[MASTER] EN-1714 Add group code not valid presentation error

fedelocci:
add behavior when on focus code input


entando-service:
All is well

PR #464, opened 2018-09-06
EN-1473 Fix fragments filter functionality

otarza:
This PR fixes only Widget type filter.
It also disables Plugin filter until API endpoint is implemented for plugins.


entando-service:
All is well

PR #492, opened 2018-11-14
EN-2220 Implemented initial mock API for DE plugins.

otarza:
https://jira.entando.org/browse/EN-2220


entando-service:
All is well

PR #505, opened 2018-12-03
EN-2249 Implementing Marketplace sidebar filter.

otarza:
Ready to be reviewed.

https://jira.entando.org/browse/EN-2249


entando-service:
All is well

entando-service:
All is well

entando-service:
All is well

sass/digital-exchange/common/SidebarFilter.scss
```diff
@@ -0,0 +1,16 @@
+.SidebarFilter {
+  text-align: left;
+}
+
+
+.SidebarFilter__title {
```
gidesan: I'd suggest `&__title` inside `.SidebarFilter` (sample reference: like https://github.com/entando/app-builder/blob/master/sass/categories/common/CategoryTreeSelector.scss)

sass/digital-exchange/common/SidebarFilter.scss
```diff
@@ -0,0 +1,16 @@
...
+  color: #4D5258;
+  font-family: Open Sans;
+  font-style: normal;
+  font-weight: bold;
+  line-height: normal;
+  font-size: 12px;
+  background-color: #E7E7E7;
```
gidesan: We should use SASS variables for colors: it improves readability, consistency and it's useful if we want to support theming in DE (@nicpuddu is it planned AFAYK?)

entando-service:
All is well

entando-service:
All is well

entando-service:
All is well

entando-service:
All is well

sass/digital-exchange/common/SidebarFilter.scss
```diff
@@ -0,0 +1,17 @@
+.SidebarFilter {
+  text-align: left;
+}
+
+.SidebarFilter {
+  &__title {
```
gidesan: you can do 
```
.SidebarFilter {
  text-align: left;
  
  &__title {
```

src/ui/digital-exchange/MarketplaceFilter.js
```diff
@@ -0,0 +1,39 @@
...
+    this.props.onWillMount();
+  }
+
+  render() {
+    const options = this.props.digitalExchangeMarketplaces.map(marketplace => (
+      { label: marketplace, value: marketplace }
+    ));
+
+    return (
+      <SidebarFilter title="Marketplace">
```
gidesan: I'm afraid you will also get an unwanted tooltip using `title` as prop name. Do we need that?

otarza: @gidesan I don't see a tooltip, where do you see it?

entando-service:
All is well

src/ui/digital-exchange/MarketplaceFilter.js
gidesan: if no tooltip that's fine, I can resolve the conversation

entando-service:
All is well

src/ui/digital-exchange/components/list/ComponentListPage.js
```diff
@@ -0,0 +1,33 @@
...
+
+
+const ComponentListPage = () =>
+  (
+    <InternalPage className="ComponentListPage">
+      <Grid>
+        <Row>
+          <Col md={12}>
+            <PageTitle
+              titleId="digital-exchange.component.list.title"
```
gidesan: seems that translation for this id is missing in both en and it locales

src/ui/digital-exchange/components/list/ComponentListPage.js
```diff
@@ -0,0 +1,33 @@
...
+              helpId="digital-exchange.component.help"
```
gidesan: seems that translation for this id is missing in both en and it locales

sass/digital-exchange/common/SidebarFilter.scss
gidesan: Sorry Otar, I still see `.SidebarFilter` twice. Am I missing something?

src/ui/digital-exchange/components/list/ComponentListContent.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import { Row } from 'patternfly-react';
+
+import ComponentListContainer from 'ui/digital-exchange/components/list/ComponentListContainer';
+
+const ComponentListContent = () => (
```
gidesan: it sounds a bit confusing to me: why a `ComponentListContent` contains a `ComponentListContainer`?

otarza: @gidesan hierarchy here looks like this: 

`ComponentListContent ` > `ComponentListContainer` > `ComponentList`

I understand your concern but `ComponentListContent` is considered as more broad component, which can encapsulate everything(content) related to component list while `ComponentListContainer` is direct parent of `ComponentList` and also connects component with redux store. 

This kind of hierarchy and logic is applied to  other parts of the project, so I decided to follow that. You can check `FragmentListContent` for an example. 

While hierarchy seems correct to me I think we can improve the `ComponentListContent` name to be more intuitive.

gidesan: But in your example a `FragmentListContent ` contains `FragmentSearchFormContainer`, not a `FragmentListContainer`. And this could make sense, because are content and container related to different components.

otarza: @gidesan it also contains `FragmentListTableContainer` which is same level as `ComponentListContainer`, it just has Table in the name to indicate that it is a table.

nicpuddu: Ok guys, you are both right in this instance: you may end up with a component that is containing several other components, which might be containers used to connect the child components to the redux store, just as @otarza pointed out. In this instance though it seems that the only purpose of this component is currently literally containing one container, which kinda looks like a poltergeist anti-pattern. If we are literally one PR away from adding additional containers/components in the body of this one we can leave it there, otherwise this is going to be an unnecessary wrapper to future proof something that might never exist.

please act accordingly

otarza: Thanks @nicpuddu, 

@gidesan let's remove for now and we will add it in the future if we need it.

entando-service:
All is well

src/ui/digital-exchange/MarketplaceFilterContainer.js
```diff
@@ -0,0 +1,37 @@
...
+import MarketplaceFilter from 'ui/digital-exchange/MarketplaceFilter';
+
+const FIELD_OPERATORS = {
+  marketplace: FILTER_OPERATORS.LIKE,
+};
+
+export const mapDispatchToProps = dispatch => ({
+  onWillMount: () => (dispatch(fetchDEMarketplaces())),
+  onChange: (eventOrValue) => {
+    if (Object.prototype.hasOwnProperty.call(eventOrValue, 'marketplaces')) {
```
gidesan: we can make it simpler with
```
const { marketplaces } = eventOrValue;
if (marketplace) {
```

entando-service:
All is well

PR #508, opened 2018-12-10
EN-2272 adding initial implementation of switching grid and list views

otarza:
Work in progress, do not merge yet!

https://jira.entando.org/browse/EN-2272

### DIFF
--- a/src/state/digital-exchange/components/actions.js
+++ b/src/state/digital-exchange/components/actions.js
@@ -1,4 +1,8 @@
-import { SET_DE_COMPONENTS, SET_SELECTED_DE_COMPONENT } from 'state/digital-exchange/components/types';
+import {
+  SET_DE_COMPONENTS,
+  SET_SELECTED_DE_COMPONENT,
+  SET_DE_COMPONENT_LIST_VIEW_MODE,
+} from 'state/digital-exchange/components/types';
 import { addErrors } from '@entando/messages';
 import { toggleLoading } from 'state/loading/actions';
 
@@ -17,6 +21,13 @@ export const setDEComponents = digitalExchangeComponents => ({
   type: SET_DE_COMPONENTS,
   payload: {
     digitalExchangeComponents,
+  },
+});
+
+export const setDEComponentsListViewMode = listViewMode => ({
+  type: SET_DE_COMPONENT_LIST_VIEW_MODE,
+  payload: {
+    listViewMode,
   },
 });
 
--- a/src/state/digital-exchange/components/reducer.js
+++ b/src/state/digital-exchange/components/reducer.js
@@ -1,6 +1,10 @@
 
 import { combineReducers } from 'redux';
-import { SET_SELECTED_DE_COMPONENT, SET_DE_COMPONENTS } from 'state/digital-exchange/components/types';
+import {
+  SET_SELECTED_DE_COMPONENT,
+  SET_DE_COMPONENTS,
+  SET_DE_COMPONENT_LIST_VIEW_MODE,
+} from 'state/digital-exchange/components/types';
 
 
 const selected = (state = {}, action = {}) => {
@@ -21,7 +25,17 @@ const list = (state = [], action = {}) => {
   }
 };
 
+const listViewMode = (state = 'list-view', action = {}) => {
+  switch (action.type) {
+    case SET_DE_COMPONENT_LIST_VIEW_MODE: {
+      return action.payload.listViewMode;
+    }
+    default: return state;
+  }
+};
+
 export default combineReducers({
   selected,
   list,
+  listViewMode,
 });
--- a/src/ui/digital-exchange/components/common/ComponentListViewModeSwitcherContainer.js
+++ b/src/ui/digital-exchange/components/common/ComponentListViewModeSwitcherContainer.js
@@ -1,0 +1,37 @@
+import React from 'react';
+import { connect } from 'react-redux';
+import { setDEComponentsListViewMode } from 'state/digital-exchange/components/actions';
+import PropTypes from 'prop-types';
+
+
+const ComponentListViewModeSwitcher = ({ changeViewMode }) => {
+  function handleClick(e) {
+    e.preventDefault();
+    changeViewMode('grid-view');
+  }
+
+  return (
+    <div className="ComponentListViewModeSwitcher">
+      <button onClick={handleClick}>
+        Switch to Grid view
+      </button>
+    </div>
+  );
+};
+
+
+ComponentListViewModeSwitcher.propTypes = {
+  changeViewMode: PropTypes.func.isRequired,
+};
+
+export const mapDispatchToProps = dispatch => ({
+  changeViewMode: (viewMode) => {
+    dispatch(setDEComponentsListViewMode(viewMode));
+  },
+});
+
+
+const ComponentListViewModeSwitcherContainer =
+  connect(null, mapDispatchToProps)(ComponentListViewModeSwitcher);
+
+export default ComponentListViewModeSwitcherContainer;
--- a/src/ui/digital-exchange/components/list/ComponentList.js
+++ b/src/ui/digital-exchange/components/list/ComponentList.js
@@ -7,7 +7,7 @@ class ComponentList extends Component {
     this.props.onWillMount();
   }
 
-  renderComponentList() {
+  renderComponentListView() {
     return (
       this.props.digitalExchangeComponents.map(component => (
         <div key={component.id}>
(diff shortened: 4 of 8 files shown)


entando-service:
All is well

entando-service:
All is well

entando-service:
All is well

src/locales/it.js
```diff
@@ -642,6 +642,10 @@ export default {
     'digital-exchange.component.help': 'Entando Digital Exchange è un marketplace da cui l\'utente può installare Widget, Componenti, Frammenti, Modelli di Pagina, Modeli di Contenuto e Tipi di Contenuto',
     'digital-exchange.sidebar.marketplace-filter-title': 'Marketplace',
     'digital-exchange.sidebar.category-filter-title': 'Categorie',
+    'digital-exchange.components.latest-version': 'Ultima versione',
```
gidesan: same as in en.js

src/state/digital-exchange/components/reducer.js
```diff
@@ -21,7 +25,17 @@ const list = (state = [], action = {}) => {
   }
 };
 
+const listViewMode = (state = 'grid-view', action = {}) => {
```
gidesan: please extract `grid-view` to a constant: it's used in more than one file

src/ui/digital-exchange/components/common/ComponentListViewModeSwitcherContainer.js
```diff
@@ -0,0 +1,62 @@
...
+import { connect } from 'react-redux';
+import { setDEComponentsListViewMode } from 'state/digital-exchange/components/actions';
+import { getDEComponentListViewMode } from 'state/digital-exchange/components/selectors';
+import PropTypes from 'prop-types';
+
+
+const ComponentListViewModeSwitcher = ({ viewMode, changeViewMode }) => {
+  function switchToGridView(e) {
+    e.preventDefault();
+    changeViewMode('grid-view');
```
gidesan: please extract grid-view to a constant: it's used in more than one file

src/ui/digital-exchange/components/list/ComponentList.js
```diff
@@ -53,6 +113,7 @@ ComponentList.defaultProps = {
   onWillMount: () => {},
   loading: false,
   digitalExchangeComponents: [],
+  listViewMode: 'gird-view',
```
gidesan: please extract grid-view to a constant: it's used in more than one file (here `gird-view` should be a typo, I guess

src/state/digital-exchange/components/actions.js
```diff
@@ -20,6 +24,13 @@ export const setDEComponents = digitalExchangeComponents => ({
   },
 });
 
+export const setDEComponentsListViewMode = listViewMode => ({
```
gidesan: typo: should be `setDEComponentListViewMode` instead of `setDEComponentsListViewMode`

src/ui/digital-exchange/common/StarRating.js
```diff
@@ -0,0 +1,25 @@
+import { range } from 'lodash';
+import React from 'react';
+import PropTypes from 'prop-types';
+import StarIcon from 'ui/digital-exchange/common/StarIcon';
+
+const StarRating = ({ rating }) => (
+  <div>
```
gidesan:
```suggestion
  const roundedRating = Math.round(rating)
  <div aria-label={roundedRating}>
```

(please double-check this suggestion)

gidesan: please add an aria-label with the rounded rating

src/state/digital-exchange/components/actions.js
```diff
@@ -20,6 +24,13 @@ export const setDEComponents = digitalExchangeComponents => ({
...
+  type: SET_DE_COMPONENT_LIST_VIEW_MODE,
+  payload: {
+    listViewMode,
```
gidesan: can we name this `componentListViewMode`?

src/ui/digital-exchange/components/common/ComponentListViewModeSwitcherContainer.js
```diff
@@ -0,0 +1,62 @@
+import React from 'react';
+import { connect } from 'react-redux';
+import { setDEComponentsListViewMode } from 'state/digital-exchange/components/actions';
+import { getDEComponentListViewMode } from 'state/digital-exchange/components/selectors';
+import PropTypes from 'prop-types';
+
+
+const ComponentListViewModeSwitcher = ({ viewMode, changeViewMode }) => {
+  function switchToGridView(e) {
```
gidesan: I don't see use of `function` in our codebase. @nicpuddu do we have any guideline about this?
Generally speaking I like functions, but in this case I'd like to keep the codebase consistent and use an arrow function with `const` or a class method

src/ui/digital-exchange/components/common/ComponentListViewModeSwitcherContainer.js
```diff
@@ -0,0 +1,62 @@
...
+    e.preventDefault();
+    changeViewMode('grid-view');
+  }
+
+  function switchToListView(e) {
+    e.preventDefault();
+    changeViewMode('list-view');
```
gidesan: please extract `list-view` to a constant as well

src/ui/digital-exchange/components/list/ComponentList.js
```diff
@@ -36,6 +95,7 @@ class ComponentList extends Component {
 ComponentList.propTypes = {
   onWillMount: PropTypes.func,
   loading: PropTypes.bool,
+  listViewMode: PropTypes.string,
```
gidesan: IMO better naming it `viewMode`, because (if I understood) it's the view mode of the component list (and so it could be either `list-view` or `grid-view`)

src/ui/digital-exchange/components/list/ComponentList.js
```diff
@@ -1,13 +1,17 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
-import { Spinner } from 'patternfly-react';
+import { Spinner, Row, Col } from 'patternfly-react';
+import moment from 'moment';
+import { formattedText } from '@entando/utils';
+import { renderInstallActions, renderComponentImageOrPlaceholder } from 'helpers/digital-exchange/components';
+import StarRating from 'ui/digital-exchange/common/StarRating';
 
 class ComponentList extends Component {
```
gidesan: What about splitting this component in two different ones? One for the list view mode and another for the grid view mode

gidesan: What about splitting this component in two different ones? One for the list view mode and another for the grid view mode

src/ui/digital-exchange/components/list/ComponentList.js
```diff
@@ -22,11 +26,66 @@ class ComponentList extends Component {
     );
   }
 
+  renderComponentGridView() {
+    const componentPairs = this.props.digitalExchangeComponents
```
gidesan: In grid view, do we have component pairs at every screen resolution?

src/helpers/digital-exchange/components.js
```diff
@@ -0,0 +1,40 @@
+import React from 'react';
+import { formattedText } from '@entando/utils';
+import { Button } from 'patternfly-react';
+
+// eslint-disable-next-line import/prefer-default-export
+export const renderInstallActions = (component) => {
```
gidesan: Why export render methods instead of react components?

src/helpers/digital-exchange/components.js
```diff
@@ -0,0 +1,40 @@
+import React from 'react';
+import { formattedText } from '@entando/utils';
+import { Button } from 'patternfly-react';
+
+// eslint-disable-next-line import/prefer-default-export
```
nicpuddu: this is also only necessary if there is just one named export within a file. please remove the rule disabling if it is not necessary

src/ui/digital-exchange/components/common/ComponentListViewModeSwitcherContainer.js
nicpuddu: I agree, either use an anonymous arrow function or declare the function as a constant.

src/ui/digital-exchange/components/list/ComponentList.js
```diff
@@ -22,11 +26,66 @@ class ComponentList extends Component {
     );
   }
 
+  renderComponentGridView() {
```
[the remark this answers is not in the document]
nicpuddu: I agree, it makes sense to use the same component only whenever there are small changes within its render method. If we want this component to behave as a switch which will select the correct component I'm fine with that but let's avoid huge conditionals that are mimicking the correct solution

entando-service:
All is well

entando-service:
All is well

src/ui/digital-exchange/components/list/ComponentList.js
otarza: @gidesan that is not clear at this point, so I refactored it to fit any scenario, no pairs used any more.

gidesan: IMO it's far better now!

PR #512, opened 2018-12-18
EN-2272 Adding tests for DE GridView and related components

otarza:
https://jira.entando.org/browse/EN-2272


entando-service:
All is well

PR #516, opened 2019-01-14
En 2282 page settings button

gidesan:
Just a note: I wasn't able to call the page PATCH when clicking the modal save button because the PATCH stories are not merged yet. I will do it in the next US.


entando-service:
All is well

PR #533, opened 2019-03-25
EN-2428 - fixing styling issue that prevented install button click.

otarza:
https://jira.entando.org/browse/EN-2428


entando-service:
All is well

PR #534, opened 2019-03-26
EN-2326 persisting locale info into local storage.

otarza:
https://jira.entando.org/browse/EN-2326


entando-service:
All is well

PR #538, opened 2019-04-09
EN-2412 DE component uninstallation

jeffgo10:


entando-service:
All is well

src/api/digital-exchange/components.js
```diff
@@ -36,7 +43,26 @@ export const getDEComponentInstall = id => (
   makeRequest({
     uri: `/api/digitalExchange/install/${id}`,
     method: METHODS.GET,
-    mockResponse: COMPONENT_INSTALLATION_IN_PROGRESS,
+    mockResponse: COMPONENT_INSTALLATION_COMPLETED,
```
gidesan: why this change?

src/state/digital-exchange/components/reducer.js
```diff
@@ -23,25 +25,32 @@ const selected = (state = {}, action = {}) => {
...
-      if (componentIndex === -1) {
-        return state;
-      }
-      const newState = state.slice();
-      newState.splice(componentIndex, 1, {
-        ...state[componentIndex],
-        installed: true,
-      });
-      return newState;
+      return findAndSetInstall(state, action.payload.id, true);
```
gidesan: What about three methods? A `markComponentInstallStatus` with a boolean argument, plus, `markComponentAsInstalled` (calls markComponentInstallStatus with true) `markComponentAsUninstalled` (calls markComponentInstallStatus with true). Could be clearer for the reader.

src/api/digital-exchange/components.js
jeffgo10: this was changed because when using mock mode it doesn't really finish the mock installation process when you try to install. I can change it back though but I might as well add a comment to use `COMPONENT_INSTALLATION_COMPLETED` when testing installation in mock mode. would that be okay?

src/state/digital-exchange/components/reducer.js
jeffgo10: i'll see what i can do with this. thanks for the tip

jeffgo10:
changes done. please review @gidesan

entando-service:
All is well

src/api/digital-exchange/components.js
gidesan: ok then, keep your change

jeffgo10: reverted

entando-service:
All is well

PR #539, opened 2019-04-11
EN-2282 - amend toolbar buttons in Page Configuration

jeffgo10:
all the top toolbar buttons are now transparent based on figma. bottom buttons are now in a sticky toolbar right at the bottom.


entando-service:
All is well

entando-service:
All is well

PR #540, opened 2019-04-15
EN-2309 - page config toolbar scroll fix when it is in sticky mode

jeffgo10:
Sorry on finding this late. I found a bug when toolbar is in fixed/sticky mode, toolbar scrolling does not work. This was overlapped with another 'dummy' element thus making scroll wheel not working.


entando-service:
All is well

PR #541, opened 2019-04-17
EN-2309 - page config toolbar scrolling drag item fix

jeffgo10:
fix on disabled draggable item at the bottom of the toolbar view - bug spotted by Simone


entando-service:
All is well

PR #542, opened 2019-04-24
EN-2355 polling timeout notifications

jeffgo10:
Added notifications after polling timeout for unfinished DE component installation and uninstallation. This will also provide extra action button for such state, such as "re-check" to recheck installation/uninstallation status and also "retry" for failed installation/uninstallation. Feel free to make corrections on the UX of the follow-up action controls.


entando-service:
All is well

nicpuddu:
> There are some usages of `formattedText` we should avoid using.
> My opinion is approving this PR including those `formattedText` calls but creating a new technical debt user story about `formattedText` removal that will cover:
> 
> * removing formattedText usages
> * removing formattedText from `@entando/utils`
> * using `injectIntl` HOC from `react-intl` where possible (components)
> * avoid passing translated strings to `addErrors` or `addToasts` actions (like in this case) and give the translations responsibility to error and toast components
> 
> @nicpuddu agree?

fine with me! let's create the story, though

gidesan:
Created https://jira.entando.org/browse/EN-2503
Will approve and merge the PR then

PR #545, opened 2019-05-07
EN-2411 - dependency update - plugins store config change to callbacks

jeffgo10:
this is a fix for plugin store config that shows a callback error when you install jpkiebpm plugin


entando-service:
All is well

PR #548, opened 2019-07-03
adding keycloak integration to app-builder

sergiofilhowz:


entando-service:

entando-service:
All is well

entando-service:
All is well

entando-service:
All is well

config/webpack.config.dev.js
```diff
@@ -31,6 +31,21 @@ const mapToFolder = (dependencies, folder) =>
...
+ */
+const lintDisabled = process.env.DISABLE_LINT === 'true';
+const lintOptions = lintDisabled ? [] : [{
+  options: {
+    formatter: eslintFormatter,
+    eslintPath: require.resolve('eslint'),
+  },
+  loader: require.resolve('eslint-loader'),
+}];
+
```
nicpuddu: why are we giving the option to disable the linter?

sergiofilhowz: Hi Nicola, let me give you an example:

While I was developing I was receiving some weird errors on the Context provider (that was because of the version of react-dom), The error logs couldn't help a bit, so I had to filter what was causing so I had to remove some elements one by one to identify what element was causing this. Every time I removed one element, I had to work on on unused variables, unused imports, and a lot of things on the eslint, all of them unnecessary to the work I was doing, because none of the changes I wasn't going to commit. I can say that I lost (literally lost) 3 hours in total doing fixes on the code to be eslint compliant. So, with an option to disable temporarily this linter will help on those situations.

I literally lost 3 hours because all the changes I did to identify the problem resulted on a single line commit on `package.json`

nicpuddu: hi Sergio, you could have just added the disable eslint comment in the file.

```
/* eslint-disable */

alert('foo');

/* eslint-enable */
```

this is totally unnecessary since it just disable eslint globally on the whole project which should not be needed

sergiofilhowz: Sometimes I had to remove a large block and makes several lines unused. Anyway, I'll do what you want since you are saying it's "totally unnecessary". I'm going to fix this.

entando-service:
All is well

PR #551, opened 2019-07-28
router refactoring

nicpuddu:


entando-service:

entando-service:

entando-service:

gidesan:
retest this please

entando-service:

entando-service:
All is well

src/ui/app/App.js
```diff
@@ -174,124 +178,188 @@ import CreateFolderPage from 'ui/file-browser/add/CreateFolderPage';
...
+  } else if (!username && currentRoute !== ROUTE_HOME) {
+    return <Redirect to={ROUTE_HOME} />;
   }
 
   return (
     <Fragment>
       <ToastsContainer />
       {!keycloak.enabled || keycloak.authenticated
-        ? getRouteComponent(route, keycloak)
+        ? getRouteComponent(keycloak)
```
gidesan: the keycloak flag seems to not be used in the `getRouteComponent` method

Apps.md
```diff
@@ -0,0 +1,80 @@
...
+several different items need to be exported by the App so that they can be consumed by the app builder.
+
+Each paragraph reports in angle brackets the name of the object that the App itself need exporting
+
+## ID <id>
+the unique identifier for the app:
+
+i.e. `cms`
+
+## Routing
```
gidesan: is the routing paragraph still needed after the removal of @entando/router?

gidesan:
I don't want to spend too much time on this story because it needs to be quickly merged, but before merging I want to share some notes:

- having some “selector creators” instead of normal selector (to overcome the removal of route parameter in redux state) can be misleading
- as an alternative, we could have used react-router-redux and have a better retro-compatibility with our code, avoid selector creators and keep using removed selectors like getParams
- we could also have kept gotoRoute with a different implementation (a simple call to history.push)

entando-service:
All is well

Apps.md
nicpuddu: I'll be taking care of this later

gidesan: cool, thanks

PR #553, opened 2019-08-15
EN6-34

sergiofilhowz:
adding digital exchange as a separate service and fixes on keycloak session token expiration


entando-service:
All is well

sergiofilhowz:
I needed to change the endpoints because Digital Exchange is now using the `entando-component-manager` (so called k8s-service). I had to create a PR on `@entando/apimanager` to enable custom subpaths to access the service's url inside kubernetes cluster (`/digital-exchange`).

sergiofilhowz:
https://github.com/entando/frontend-libraries/pull/35

entando-service:
All is well

Kerruba:
> Guys, I'm afraid there's some entropy increase.
> 
>     * PATH_PREFIX is not exactly a prefix, as it comes after DOMAIN in the final request url
> 
>     * PATH_PREFIX goes into the apiManager store section and it can be overwritten by a `path` (why not pathPrefix then?) property in the `makeRequest` object
> 
>     * in `makeRequest` calls is not exactly easy to understand what `uri` is, what `path` is and who comes first in the final request url
> 
>     * (and maybe DOMAIN is not exactly about a domain)
> 
the final HTTP request path will be something like this
`{DOMAIN:/}/{PATH_PREFIX:entando-de-app}/{URI:/api/widgets}`

basically PATH_PREFIX is really the service we want to create the request against (staying on the same domain). In Entando6, that service could be entando-core, the digital-exchange, the config-service and potentially other services that are exposed in the same domain.

Possible solutions to this are:
1. use complete URIs specifying the path-prefix (e.g. api/widgets -> entando-de-app/api/widgets)
2. for the options passed to the `makeRequest` method, we can specify two different properties: `entandoUri` are uris built taking the entando-de-app as prefix, `uri`(or `genericUri` are instead complete uri (relative or absolute shouldn't matter)


> On top of that, are we sure we want to remove the `/api` prefix that should be common to other Entando APIs?

For this I think from the pov of REST, having the `/api` endpoint is not really meaningful, as the endpoint should return different results based on the `Accept` header. Thus if you set `Accept: text/html` the html page is returned and if you set `Accept: application/json` the json is returned.

> As a quick fix, I'd rename `PATH_PREFIX` and `path` to `URI_PREFIX` and `uriPrefix` at least.

This could be a simple yet effective solution

joewhite101:
The approach recommended by @Kerruba seems reasonable to me. @gidesan do you have concerns on that structure?

Kerruba:
@joewhite101 which of the two approaches?

gidesan:
@joewhite101 my main concern is about keeping I want to keep backwards compatibility so we don't create regressions in other apps using `apiManager`

Kerruba:
Personally I prefer the approach 1, and we can always convert the requests to entando backend using a template string in the uri, e.g. `{process.env.entando_url}/api/...` 

I think this is the clearest and cleanest approach

gidesan:
+1 for the approach 1, it's a bit more verbose but clearer and easier to understand

nicpuddu:
I agree, I would remove the option to pass it to makeRequest

entando-service:
All is well

entando-service:
All is well

entando-service:
All is well

PR #555, opened 2019-10-02
EN6-72 rebased Docker image to use standard Openshift Nginx image

ampie:
These changes are required for SME to allow the AppBuilder image to be deplyable to Openshift. The previous version of the AppBuilder image was incompatible with Openshift due to a requirement to run as root, which Openshift disallows for security reason.


entando-service:
All is well

joewhite101 (approved):
Looks good.

PR #562, opened 2019-10-21
EN6-132 Update selected page status when page widgets change

raxkaynan:
Fixes EN6-132 issue: Cannot publish twice from page configuration


entando-service:
All is well

PR #566, opened 2019-11-04
EN6-145 search box for code editors

jeffgo10:
This is to add search feature for most code editors (specifically using 'CodeMirror' component). I only activated existing addons (found it here, https://codemirror.net/demo/search.html).


entando-service:
All is well

PR #567, opened 2019-11-04
En6 154 integrations menu

ichalagashvili:
Removed integrations item from menu


entando-service:
All is well

entando-service:
All is well

PR #569, opened 2019-11-05
EN6-146 remove file browse filter in file browser for upload

jeffgo10:
literally just remove acceptfile prop. nothing more, nothing less. :D


entando-service:
All is well

PR #570, opened 2019-11-05
EN6-147 my profile page fix

jeffgo10:
It was the 'pure' option in redux connect function that causes the page to go haywire - infinite loops of checks of state values most especially the `matchElement` validation that is found on `newPasswordConfirm`. If there's a matchElement validation, better avoid turning on `pure` option.


entando-service:
All is well

PR #574, opened 2019-11-07
EN6-152 deduplicating dependencies

jeffgo10:
resolving duplicates by upgrading some key dependencies and adding resolve.alias to some dependencies that can't be resolve on updating. Also including apimanager from 4.1.0 to 4.2.0


entando-service:
All is well

PR #581, opened 2019-11-25
EN6-177 key requirement error fix for integrated apps

jeffgo10:
Quick fix by adding missing key attribute


entando-service:
All is well

PR #582, opened 2019-11-28
EN6-172 dispatch prop onWillMount not found error fix

maitojepoy:
prop onWillMount not found error when going to user list


entando-service:
Can one of the admins verify this patch?

PR #583, opened 2019-11-28
EN6-172 dispatch prop onWillMount not found error fix

jeffgo10:
prop 'onWillMount' attempts to invoke but not found on user list. this only happens when it is integrated with entando app such as CMS


entando-service:
All is well

PR #585, opened 2019-12-02
ENG-7 Activate a spinner while in loading state

ichalagashvili:
This commit includes spinner activation and no-scroll issue resolve.
1. Spinner activation was accessing the spinner status incorrectly from the state. Also, toggleLoading was never called when there are no Digital Exchanges available, as API is returning 304 code without any JSON, so frontend fetching PROMISE outputs error (no JSON returned) instead of resolving. That's why I put toggleLoading in error too.
2. I tested while no-scroll removed and all worked like a charm and smoothly. So, I think this way is better.


entando-service:
All is well

PR #587, opened 2019-12-05
EN6-171 entando pages updated to 2.2

jeffgo10:
in order to use the mobile/tablet compatible login screen from @entando/pages


entando-service:
All is well

PR #589, opened 2019-12-11
EN6-134 login background flicker on page open

jeffgo10:
this is keycloak related and solution is to need to render keycloak before anything else including apimanager. So solution is to refactor api manager into a component (like what we did in entando-pwa). Also major refactoring on auth components for default and keycloak.


entando-service:
All is well

gidesan:
How the refactor to a component helped with solving the background flicker?

jeffgo10:
> How the refactor to a component helped with solving the background flicker?

Before, the api manager executes first before keycloak initiates with verifying tokens, this result to a render starts to show portion of login screen without even waiting for keycloak to verify. at the midst of rendering that keycloak auth is verified, it quickly shifted to render the dashboard page. 

Back at entando-pwa, we used `authInitialized` from keycloak HOC as a flag to render everything else once keycloak has done verifying. The solution I thought was basically to just let keycloak or whatever auth execute first before apimanager does. That's why I thought this needs to refactor to component(s). You can see now that in index.js, <AuthProvider> (or Keycloak/default auth) is at the top before <ApiManager>.

entando-service:
All is well

PR #591, opened 2019-12-11
ENG-30 provide feedback if DEs or corresponding components are not presented

ichalagashvili:
- Text visual feedbacks are provided

### DIFF
--- a/src/state/digital-exchange/components/actions.js
+++ b/src/state/digital-exchange/components/actions.js
@@ -1,6 +1,8 @@
 import {
   addErrors,
   addToast,
+  clearErrors,
+  TOAST_ERROR,
   TOAST_WARNING,
 } from '@entando/messages';
 import { formattedText } from '@entando/utils';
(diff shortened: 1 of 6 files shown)


entando-service:
All is well

src/state/digital-exchange/components/actions.js
```diff
@@ -268,11 +270,13 @@ export const fetchDEComponents = (paginationMetadata = {
           dispatch(setPage(data.metaData));
         } else {
           dispatch(addErrors(data.errors.map(err => err.message)));
+          data.errors.forEach(err => dispatch(addToast(err.message, TOAST_ERROR)));
```
gidesan: I remembered we preferred to have one toast for multiple errors. @nicpuddu can you please confirm that?

nicpuddu: this only makes sense (multiple toasts) if the errors are not related to one another... if they are all tied together it should be a single toast.

nicpuddu: on top of that the errors should be human readable and not technical mumbo jumbo, remember that.

ichalagashvili: DE components error handling is another story which is blocked by BE and will be implemented in the future, so I think we can accept this atm.

gidesan: @ichalagashvili is there any update on that?

ichalagashvili: So, I investigated it and errors messages here are independent. So, multiple toasts make sense. (DE component A can be failed for reason X and DE component B can be failed for another reason Y). But, in general, maybe the FETCHING logic for DE components can be changed, or at least the error reporting logic. See: https://jira.entando.org/browse/ENG-93

gidesan: there are lots of DE changes planned, we can accept the PR as it is atm

PR #592, opened 2019-12-12
ENG-58 page delete option disabled fix in page search results

jeffgo10:
also added removal of page item in search result list after prompting to remove page


entando-service:
All is well

PR #596, opened 2019-12-18
ENG-36 catch unathorized error

ichalagashvili:
`login` function responds with 401 unauthorized error from API, which does not go to `then` but to `catch`, where we did not have any logic to handle it, due to it UX was not indicating any error and the user thought he was stuck on the page and screen was `frozen`.


entando-service:
All is well

PR #604, opened 2020-01-14
ENG-97 fix app-builder vulnerabilities

ichalagashvili:


entando-service:
All is well

gidesan:
does it fix that react-script issue you mentioned today as well?

ichalagashvili:
> does it fix that react-script issue you mentioned today as well?

@gidesan this one is for `app-builder` all is good here, you can approve. The problematic one is `entando-cms`.

PR #606, opened 2020-01-15
EN-59 other i18n fixes

gidesan:
use "tipo di widget" to translate "widget type" as it was already used that way


entando-service:
All is well

PR #609, opened 2020-01-22
EN6-136 cms widgets

ichalagashvili:
Import cms widgets into app-builder


entando-service:
All is well

src/ui/widgets/config/WidgetConfigForm.js
```diff
@@ -1,19 +1,45 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { intlShape } from 'react-intl';
 import Form from 'react-jsonschema-form';
+import apps from 'entando-apps';
 
-const WidgetConfigForm = ({ onSubmit, widgetConfig }) =>
-  (widgetConfig && widgetConfig.schema ? (
+const widgetForms = apps.reduce((obj, app) => ({
```
gidesan: This shouldn't be coupled with CMS. It should be an integration point for every kind of appbuilder app

entando-service:
All is well

entando-service:
All is well

ichalagashvili:
- With current approach if two different apps have widget config form with same ids, then one will overwrite another. To fix that I think its good if we pass `app.id` too inside `putPageWidget` API call, then during reading the widget config from BE, we will be able to distinguish which config belongs to which app's widget form. @gidesan

gidesan:
> * With current approach if two different apps have widget config form with same ids, then one will overwrite another. To fix that I think its good if we pass `app.id` too inside `putPageWidget` API call, then during reading the widget config from BE, we will be able to distinguish which config belongs to which app's widget form. @gidesan

Makes sense!

PR #612, opened 2020-01-28
ENG-115

ichalagashvili:
- Add *{pure: false}* in some containers where it was missing
- Change implementation design of `validatePreviewErrors ` function
- Update corresponding tests


entando-service:
All is well

PR #621, opened 2020-02-05
ENG-109 Use isURL validator to validate DOMAIN env variable

raxkaynan:
Fixes issue on paths with improper protocols.


entando-service:
All is well

PR #622, opened 2020-02-06
ENG-148 fix routing when PUBLIC_URL is not empty...

gidesan:
...downgrade node-sass 4.13.0 to avoid strange behaviors of appbuilder


entando-service:
All is well

PR #628, opened 2020-02-10
Eng 108 widget config mfe

gidesan:
note: shadow dom support is not included as it should not be useful for widget config MFE


entando-service:
All is well

entando-service:
All is well

PR #1141, opened 2021-10-18
ENG-2687: keep tabs components mounted to fix validation issues

gReis89:
`mountOnEnter` and `unmountOnExit` props were keeping the unselected tabs unmounted, and unmounted component doesn't validate itself through redux-forms. Removing these props makes the fields hidden but mounted, so required fields on hidden tabs will start to validate as we want.


gReis89:
closing this PR as it was branched out from `master` and not `develop`

PR #1143, opened 2021-10-18
ENG-2687: validate Page title fields even when they are on a hidden tab

gReis89:
mountOnEnter and unmountOnExit props were keeping the unselected tabs unmounted, and unmounted component doesn't validate itself through redux-forms. Removing these props makes the fields hidden but mounted, so required fields on hidden tabs will start to validate as we want.

### DIFF
--- a/src/ui/pages/common/SeoInfo.js
+++ b/src/ui/pages/common/SeoInfo.js
@@ -23,7 +23,7 @@ const SeoInfo = ({
   formId,
 }) =>
   (languages && languages.length ? (
-    <Tabs id="basic-tabs" defaultActiveKey={0} className="SeoInfo" mountOnEnter unmountOnExit>
+    <Tabs id="basic-tabs" defaultActiveKey={0} className="SeoInfo">
       {
       languages.map((lang, i) => (
         <Tab key={lang.code} eventKey={i} title={`${lang.code.toUpperCase()}${i === 0 ? '*' : ''}`} >
--- a/src/ui/pages/common/SeoInfo.js
+++ b/src/ui/pages/common/SeoInfo.js
@@ -39,7 +39,7 @@ const SeoInfo = ({
     if (touchedTabs.includes(key) || defaultLang.code === key) {
       return [required, maxLength70];
     }
-    return false;
+    return [];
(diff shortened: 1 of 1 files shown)


raxkaynan (changes requested):
After doing a quick test, I found one problem: the page can't be saved unless you fill in the title for every other language. AFAIK, you should be able to save the page if you're in the default language tab and have filled in the required fields.

gReis89:
@raxkaynan good catch! So it doesn't make more sense making the other language fields optional?

gReis89:
Added a logic to dynamically mark the Title field as `required` accordingly to whether or not the users have accessed another lang tab, as per @nicpuddu suggestion.

src/ui/pages/common/SeoInfo.js
```diff
@@ -21,12 +21,32 @@ const SeoInfo = ({
...
+    if (!touchedTabs.includes(key)) {
+      setTouchedTabs([...touchedTabs, key]);
+    }
+  };
+
+  const createRequiredArray = React.useCallback((key) => {
+    if (touchedTabs.includes(key) || defaultLang.code === key) {
+      return [required, maxLength70];
+    }
+    return false;
```
raxkaynan: This produces a console warning stating "Invalid prop supplied to Field". It should probably return an empty array here instead of `false`.

raxkaynan (changes requested):
The new changes seem to prevent the user from saving the page if another language tab has been "touched" but not filled out, even when the user goes back to the default tab. The previous behavior was that the save button is disabled if the "active" tab is a non-default language tab and its required title is empty.

@nicpuddu Is the newly introduced logic the correct behavior?

gReis89:
This is the agreed behaviour @raxkaynan 
once users acknowledge a tab it become required and users must fill it before proceed. The autocomplete feature will only be applied if users doesn't touch the tab

src/ui/pages/common/SeoInfo.js
gReis89: fixed!

raxkaynan (approved):

PR #1149, opened 2021-10-19
ENG-2545 Change fragment code validation to accept the dash character

raxkaynan:


src/ui/fragments/common/FragmentForm.js
```diff
@@ -18,6 +18,15 @@ import {
 
 const maxLength50 = maxLength(50);
 
+const fragmentCode = value => (
+  value && /^[0-9a-zA-Z_\-.]+$/i.test(value) ?
```
gReis89, line 22: if `value` is `0`, wouldn't it brake this logic, since `0` is a falsy value?

raxkaynan: Had to double-check it to be sure 😄. But no it doesn't since `value` is a string, not a number.

gReis89 (approved):

user-54dbb57a:
Rebased to the latest develop to get the latest pipelines updates

raxkaynan:
Closed since ticket has been rejected.

PR #1152, opened 2021-10-21
ENG-2677 Update jenkins and install script to be able to specify cms …

raxkaynan:
…version


nicpuddu (approved):
fine with me, although this is making the flag mandatory.

raxkaynan:
> fine with me, although this is making the flag mandatory.

Not sure how it's making it mandatory. There's a condition there to include the version only when it's available.

nicpuddu:
> > fine with me, although this is making the flag mandatory.
> 
> Not sure how it's making it mandatory. There's a condition there to include the version only when it's available.

that's not the best practice to do that though Rax
https://github.com/tj/commander.js/#more-configuration-1

PR #1158, opened 2021-10-26
ENG-2708 Change action and state for dashboard pages

raxkaynan:
Prevents delayed search pages update when going from dashboard to page management page, which causes conflict.


jeffgo10 (approved):
looks good

PR #1159, opened 2021-10-26
ENG-2717: fixing modal and content edit page

gReis89:


src/state/edit-content/actions.js
```diff
@@ -1,7 +1,7 @@
 import { addErrors, addToast, clearErrors, TOAST_ERROR } from '@entando/messages';
 import { initialize, formValueSelector, change } from 'redux-form';
 import moment from 'moment';
-import { pickBy, isObject, cloneDeep } from 'lodash';
+import _, { pickBy, isObject, cloneDeep } from 'lodash';
```
jeffgo10: is there a reason you used `_` instead putting `isEmpty` inside the bracket (e.g. `{ isEmpty, pickBy, etc... }`?

jeffgo10 (changes requested):
just one question...

src/state/edit-content/actions.js
gReis89: ops, I did not notice we already had the import in the file, I used the IDE autocomplete feature. I'm going to change it

jeffgo10 (approved):

PR #1160, opened 2021-10-27
ENG-2752: enforce ua-parser-js version

gReis89:
Although the ua-parser-js lib has already released fixes for the affected versions, we could still be affected if one of our dependencies tries to directly target the compromised versions.

`npm-force-resolutions` is required to npm understand how to implement `resolutions`  as it is not a built in npm feature.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,23869 @@
 {
   "name": "app-builder",
   "version": "6.4.0-SNAPSHOT",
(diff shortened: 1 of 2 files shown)


raxkaynan (approved):

gReis89:
This PR is going to be closed due to the discussion in the below ticket:
https://entando.myjetbrains.com/youtrack/issue/ENG-2751

PR #1162, opened 2021-11-04
ENG-2710 Integrate Page Template form with Formik

jeffgo10:


src/state/page-templates/helpers.js
```diff
@@ -84,3 +83,217 @@ export const validatePageTemplate = (pageTemplate) => {
...
+    const col = getNextCol(frames, matrix, bounds, xCur);
+    xCur = col.x2 + 1;
+    col.parentKey = getCellKey(bounds);
+    cols[col.key] = col;
+  }
+  return cols;
+};
+
+
+const objectValues = object => Object.keys(object).map(prop => object[prop]);
```
raxkaynan, line 206: There's probably no need for this function. You can just use `Object.values()`.

raxkaynan (changes requested):
Looks great. Just one minor change.

raxkaynan (approved):

PR #1168, opened 2021-11-09
ENG-2857 resolve disappearance of source code input value + amending CMS's Textarea Input component to AB

jeffgo10:


src/ui/common/rich-text-editor/RichTextEditor.js
```diff
@@ -211,6 +211,12 @@ class RichTextEditor extends Component {
       this.quill.clipboard.dangerouslyPasteHTML(html);
     }
     this.txtArea.style.display = this.txtArea.style.display === 'none' ? '' : 'none';
+    const parentTxtarea = this.txtArea.closest('.ql-srccode');
+    if (parentTxtarea.classList.contains('active')) {
```
raxkaynan: You can just use [`classList.toggle()`](https://developer.mozilla.org/en-US/docs/Web/API/Element/classList#:~:text=div.classList.toggle(%22visible%22)%3B) here.

jeffgo10: oh yeah I forgot about this. ugh. Thanks @raxkaynan

raxkaynan (approved):

PR #1171, opened 2021-11-10
ENG-2714: making Image field required

gReis89:
- implement validation logic for Image fields

### DIFF
--- a/src/helpers/attrValidation.js
+++ b/src/helpers/attrValidation.js
@@ -1,6 +1,6 @@
 import React from 'react';
 import { FormattedMessage } from 'react-intl';
-import { memoize, isNull, isBoolean } from 'lodash';
+import { memoize, isNull, isNil, isEmpty, isBoolean } from 'lodash';
 import { maxLength, minLength } from '@entando/utils';
 
 export const EMPTY_SYMBOLIC_DEST = '#!!#';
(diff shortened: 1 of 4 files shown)


jeffgo10 (approved):

src/helpers/attrValidation.js
```diff
@@ -202,3 +202,12 @@ export const linkValidate = memoize((langCode, required = false) => input => (
 export const listRequired = value => (
   !value || !value.length ? <FormattedMessage id="validateForm.required" /> : undefined
 );
+
+export const imageValidate = memoize((langCode, required = false) => input => (
+  (required && !input && !input.values
```
raxkaynan: Seems odd to me. If `input` is falsy, then does that mean `input.values` would throw an error? Also, the only condition that would allow the subsequent checks to be evaluated is for `input` to be falsy. Otherwise, it shouldn't work as expected.

raxkaynan (changes requested):
There seems to be erroneous logic in the validation. I've tested this as well to be sure. If you create a new content with the mandatory image and save it, then add a new mandatory image attribute to the content type used, when you edit the same content, you can save it even if the new mandatory image is empty.

src/helpers/attrValidation.js
gReis89: hi @raxkaynan ! I've simplified the logic, the problem is that the Image Field has inner fields bound to it, and the same validation function is being used to the inner fields as well.. so I had to create an awkward logic to make it work.. but now, I have break it down into different blocks, and now the issues seems to be resolved.

src/helpers/attrValidation.js
```diff
@@ -202,3 +202,16 @@ export const linkValidate = memoize((langCode, required = false) => input => (
...
   !value || !value.length ? <FormattedMessage id="validateForm.required" /> : undefined
 );
+
+export const imageValidate = memoize(langCode => (input) => {
+  let condition = false;
+  if (input && input.values) {
+    // image object
+    condition = !isEmpty(input.values[langCode]);
+  }
+  if (input[langCode]) {
```
raxkaynan, line 212: Should this be `input.values[langCode]` instead?

raxkaynan (changes requested):

raxkaynan (approved):

PR #1173, opened 2021-11-11
ENG-2808 migrate from react-form to formik for Fragment forms

jeffgo10:
also included in this PR is replacing the `<Formik>` i have used in page template forms to `withFormik` HOC as agreed with the team

### DIFF
--- a/src/helpers/formikValidations.js
+++ b/src/helpers/formikValidations.js
@@ -16,7 +16,7 @@ export const validateCodeField = intl => (value, { createError, path }) => (
   value && /^[0-9a-zA-Z_.]+$/i.test(value) ?
     true :
     createError({
-      message: intl.formatMessage('validateForm.code'),
+      message: intl.formatMessage({ id: 'validateForm.code' }),
       path,
     })
 );
--- a/src/helpers/formikValidations.js
+++ b/src/helpers/formikValidations.js
@@ -5,7 +5,7 @@ export const formatMessageRequired = {
 
 export const formatMessageMaxLength = {
   id: 'validateForm.maxLength',
-  deaultMessage: 'Maximum of {max} characters only.',
+  defaultMessage: 'Maximum of {max} characters only.',
 };
 
 export const validateJson = intl => (value, yupProps) => {
--- a/src/helpers/formikValidations.js
+++ b/src/helpers/formikValidations.js
@@ -11,3 +11,12 @@ export const validateJson = (value, yupProps) => {
     });
   }
 };
+
+export const validateCodeField = intl => (value, { createError, path }) => (
+  value && /^[0-9a-zA-Z_.]+$/i.test(value) ?
+    true :
+    createError({
+      message: intl.formatMessage('validateForm.code'),
+      path,
+    })
+);
--- a/src/helpers/formikValidations.js
+++ b/src/helpers/formikValidations.js
@@ -1,4 +1,13 @@
-// eslint-disable-next-line import/prefer-default-export
+export const formatMessageRequired = {
+  id: 'validateForm.required',
+  defaultMessage: 'Required',
+};
+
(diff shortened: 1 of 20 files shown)


src/helpers/formikValidations.js
```diff
@@ -1,13 +1,31 @@
-// eslint-disable-next-line import/prefer-default-export
-export const validateJson = (value, yupProps) => {
+export const formatMessageRequired = {
+  id: 'validateForm.required',
+  defaultMessage: 'Required',
+};
+
+export const formatMessageMaxLength = {
+  id: 'validateForm.maxLength',
+  deaultMessage: 'Maximum of {max} characters only.',
```
ichalagashvili: please fix typo

ichalagashvili: `deaultMessage`

src/locales/en.js
```diff
@@ -376,6 +376,7 @@ export default {
     'validateForm.maxValue': 'Must be at most {max}.',
     'validateForm.regex': 'Must match this regex: {regex}',
     'validateForm.email': 'Invalid email address.',
+    'validateForm.invalidJson': 'Invalid JSON format: {msg}',
```
ichalagashvili, line 379: please also update pt.js file with same key

src/state/fragments/actions.js
```diff
@@ -199,6 +206,7 @@ export const sendPostFragment = (fragment, saveType) => async (dispatch) => {
       { id: 'fragment.created' },
       TOAST_SUCCESS,
     ));
+    dispatch(clearSelectedFragment());
```
ichalagashvili: should not same `clearSelectedFragment` be dispatched also during `put` and `delete` fragment actions?

src/ui/fragments/add/AddFormContainer.js
```diff
@@ -1,22 +1,26 @@
...
 import { ConfirmCancelModalID } from 'ui/common/cancel-modal/ConfirmCancelModal';
 
+export const mapStateToProps = () => ({
+  initialValues: DEFAULT_FORM_VALUES,
+});
+
 export const mapDispatchToProps = (dispatch, { history }) => ({
   onSubmit: (fragment, saveType) => dispatch(sendPostFragment(fragment, saveType)),
-  onSave: () => { dispatch(setVisibleModal('')); dispatch(submit('fragment')); },
+  onHideCancelModal: () => { dispatch(setVisibleModal('')); },
```
ichalagashvili: is this method used after `onCancel` below is triggered and cancel is confirmed?

ichalagashvili (commented):
Very great job as usual Jeff! You have mastered Yup validations too! Just small things. 2 questions to clear it up.

src/ui/fragments/add/AddFormContainer.js
ichalagashvili: if not and its used to just hide the modal maybe naming is not really good? "cancel" should not be the part as its somehow misguiding, since we just hide the modal during save.

src/state/fragments/actions.js
jeffgo10: `clearSelectedFragment` invokes once user leaves the form (goes to another page). I used willunmount cycle. Is that okay?

ichalagashvili: yes its ok, but why do we have it here then? just wondering

jeffgo10: hmm good point. I'll remove that

src/ui/fragments/add/AddFormContainer.js
jeffgo10: I'll just rename it to `onHideModal`

src/locales/en.js
jeffgo10: done

ichalagashvili (approved):

PR #1175, opened 2021-11-15
ENG-2815 integrate group form with formik

jeffgo10:


src/ui/groups/common/GroupForm.js
```diff
@@ -25,122 +23,157 @@ const msgs = defineMessages({
...
 
-  render() {
-    const {
-      intl, invalid, submitting, mode, onChangeName, onFocus,
-      dirty, onCancel, onDiscard, onSave,
-    } = this.props;
+  useEffect(() => {
+    if (!isEdit) {
+      setFieldValue('code', values.name.replace(/\W/g, '_').toLowerCase());
+      setFieldTouched('code', true);
```
ichalagashvili, line 43: wow, does not formik automatically set `setFieldTouched` to true?

src/ui/groups/common/GroupForm.js
```diff
@@ -25,122 +23,157 @@ const msgs = defineMessages({
...
+    }
+  // eslint-disable-next-line react-hooks/exhaustive-deps
```
ichalagashvili, line 45: is it really necessary here?

ichalagashvili (approved):
approving. Just a few thoughts

src/ui/groups/common/GroupForm.js
jeffgo10: yeah. I used this because I migrated it from special dispatch (redux-form powered) function `onChangeName` in `src/ui/groups/add/AddFormContainer.js` which I just removed

src/ui/groups/common/GroupForm.js
jeffgo10: lint requires to add the used dispatch props in useEffect array parameter but I don't think it's necessary since our function does not change during runtime

PR #1180, opened 2021-11-15
ENG-2811 Re-implement user form to use formik instead of redux-form

raxkaynan:
* Rewrites user form to use formik
* Rewrites user form tests to use react-testing-library
* Rewrites user form container components to use hooks

### DIFF
--- a/src/ui/common/formik-field/SwitchInput.js
+++ b/src/ui/common/formik-field/SwitchInput.js
@@ -1,0 +1,102 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Switch, Col, ControlLabel } from 'patternfly-react';
+import { getTouchErrorByField } from 'helpers/formikUtils';
+
+const switchField = (
+  field, form, switchValue, trueValue,
+  falseValue, onToggleValue, dataTestId,
+) => {
+  const handleChange = (el, val) => {
+    const returnVal = val ? trueValue : falseValue;
+    form.setFieldValue(field.name, returnVal);
+    if (onToggleValue) {
+      onToggleValue(returnVal);
+    }
+  };
+
+  return (
+    <div data-testid={dataTestId} aria-labelledby={`switch-${field.name}`}>
+      <Switch
+        {...field}
+        value={switchValue}
+        onChange={handleChange}
+      />
+    </div>
+  );
+};
+
+const SwitchInput = ({
+  field, form, append, label, labelSize, inputSize, alignClass,
+  help, trueValue, falseValue, disabled, onToggleValue,
+}) => {
+  const switchValue = field.value === 'true' || field.value === true || field.value === trueValue;
+  const dataTestId = `${field.name}-switchField`;
+  const { touched, error } = getTouchErrorByField(field.name, form);
+
+  if (label) {
+    return (
+      <div className={`SwitchInput ${(touched && error) ? 'form-group has-error' : 'form-group'}`}>
+        <Col xs={labelSize} className={alignClass}>
+          <ControlLabel id={`switch-${field.name}`} htmlFor={field.name}>
+            {label} {help}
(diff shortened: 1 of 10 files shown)


src/ui/users/common/UserForm.js
```diff
@@ -48,227 +47,236 @@ export const renderStaticField = (field) => {
...
-  },
-  password: {
-    id: 'user.password',
-    defaultMessage: 'Password',
-  },
-  passwordConfirm: {
-    id: 'user.passwordConfirm',
-    defaultMessage: 'Confirm Password',
-  },
+const userPassCharsValid = (value, { createError, path }) => {
```
gReis89: shouldn't we place it on a validations file?

raxkaynan: Ah yes. Just noticed that there's now a `formikValidation.js` file. Will move that.

src/ui/common/formik-field/SwitchInput.js
```diff
@@ -0,0 +1,102 @@
...
+  if (label) {
+    return (
+      <div className={`SwitchInput ${(touched && error) ? 'form-group has-error' : 'form-group'}`}>
+        <Col xs={labelSize} className={alignClass}>
+          <ControlLabel id={`switch-${field.name}`} htmlFor={field.name}>
+            {label} {help}
+          </ControlLabel>
+        </Col>
+        <Col xs={inputSize || 12 - labelSize}>
+          {switchField(
```
gReis89: is there any specific reason for `switchField` not be a React component?

raxkaynan: Just copied the existing `SwitchRenderer` component to the `formik-field` folder and amended it a bit to make it work with formik. So didn't touch those parts of the code. But indeed, not sure why it was written that way. Making the changes wouldn't take long probably so will do that.

src/ui/users/add/AddFormContainer.js
```diff
@@ -12,26 +12,42 @@ import { ROUTE_USER_LIST } from 'app-init/router';
...
-    dispatch(sendPostUser(user, saveType === 'editProfile'));
-  },
-  onWillMount: () => {
-    dispatch(destroy('user'));
+const AddFormContainer = () => {
+  const dispatch = useDispatch();
+  const history = useHistory();
+  const profileTypes = useSelector(getProfileTypesOptions);
+
+  const handleMount = useCallback(() => {
```
jeffgo10, line 20: any reason why you have changed the format? we usually code this kind with `mapStateToProps` and `mapDispatchToProps`

src/ui/users/edit/EditFormContainer.js
```diff
@@ -1,34 +1,68 @@
...
-  pure: false,
-})(UserForm));
+
+const EditFormContainer = () => {
+  const dispatch = useDispatch();
+  const history = useHistory();
+  const { username } = useParams();
+  const selectedUser = useSelector(getSelectedUser);
+
+  const initialValues = useMemo(() => ({
```
jeffgo10, line 20: same question here

jeffgo10 (commented):
just having concern with the code structure highlighted

src/ui/users/add/AddFormContainer.js
raxkaynan: We're gradually introducing hooks into our codebase. And writing our components using hooks makes it more future-proof since, of course, that's react's future, which has been followed by all prominent libraries including react-redux. [Using react-redux hooks also has a few other benefits](https://stackoverflow.com/questions/59748180/should-i-use-useselector-usedispatch-instead-of-mapstatetoprops#answer-66218178). And I personally prefer that for a few other reasons (i.e. component-based, enforces more reliable testing, etc.).

jeffgo10 (approved):

PR #1186, opened 2021-11-17
ENG-2866: ensuring list mode on single content modal picker

gReis89:
- Ensuring that the content search modal will always call the contents API with `list` mode.
- Since `fetchContentsPaged` is being used through many places, I added a new parameter to define the `mode` to avoid changing it on wrong places.

### DIFF
--- a/src/state/contents/actions.js
+++ b/src/state/contents/actions.js
@@ -158,6 +158,7 @@ export const fetchContentsWithFilters = (
   quickFilterStatusParam = '',
   quickFilterOwnerGroup,
   joinGroupsToParse,
+  mode = 'full',
 ) => (dispatch, getState) => {
   const state = getState();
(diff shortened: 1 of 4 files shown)


src/state/contents/actions.js
```diff
@@ -158,6 +158,7 @@ export const fetchContentsWithFilters = (
   quickFilterStatusParam = '',
   quickFilterOwnerGroup,
   joinGroupsToParse,
+  mode = 'full',
```
jeffgo10: pls make some `const` variables (probably located in `state/contents/const.js`) regarding MODE_FULL and MODE_LIST (or whatever more specific names than those). Implement them here and areas using with `mode`.

jeffgo10 (changes requested):
just this one concern. seems `mode` has been used in various places.

jeffgo10 (approved):
LGTM 🌈

PR #1192, opened 2021-11-19
ENG-2905: protecting assets page

gReis89:
- Showing the Assets menu only if the user has explicity access to `MANAGE_RESOURCES_PERMISSION`

### DIFF
--- a/src/ui/internal-page/VerticalMenuContainer.js
+++ b/src/ui/internal-page/VerticalMenuContainer.js
@@ -51,10 +51,7 @@ const ROUTE_CMS_CONTENT_SETTINGS = '/cms/content-settings';
 const renderCmsMenuItems = (intl, history, userPermissions) => {
   const hasMenuContentsAccess = hasAccess([
     CRUD_CONTENTS_PERMISSION, VALIDATE_CONTENTS_PERMISSION], userPermissions);
-  const hasMenuAssetsAccess = hasAccess(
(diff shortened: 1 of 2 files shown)


src/ui/internal-page/VerticalMenuContainer.js
```diff
@@ -51,10 +51,7 @@ const ROUTE_CMS_CONTENT_SETTINGS = '/cms/content-settings';
 const renderCmsMenuItems = (intl, history, userPermissions) => {
   const hasMenuContentsAccess = hasAccess([
     CRUD_CONTENTS_PERMISSION, VALIDATE_CONTENTS_PERMISSION], userPermissions);
-  const hasMenuAssetsAccess = hasAccess(
-    [MANAGE_RESOURCES_PERMISSION, CRUD_CONTENTS_PERMISSION, VALIDATE_CONTENTS_PERMISSION],
```
ichalagashvili, line 55: are we sure only `MANAGE_RESOURCES_PERMISSION` can access assets?

gReis89: yeah.. MANAGE_RESOURCES_PERMISSION means `Asset Editing`.

following the AC:

```
given that a role with the following permissions has been created Access to Administration Area, Content Editing, Content Supervision
and that a user with that role has been created
and you login with that user inside AppBuilder
when you open the Contents option in the left menu
then the Asset option MUST NOT be there
```

means that people with only CRUD_CONTENTS_PERMISSION and VALIDATE_CONTENTS_PERMISSIONS, should not be able to access the assets management page

ichalagashvili (approved):

PR #1193, opened 2021-11-22
ENG-2901 search on select

ichalagashvili:


src/state/component-repository/components/selectors.js
```diff
@@ -59,12 +59,14 @@ export const getECRComponentLastInstallStatus = (state, props) => {
...
-  get(state, `componentRepositoryComponents.installation[${props.component.code}]`, '')
+export const getECRComponentInstallationStatus = createSelector(
+  [state => state.componentRepositoryComponents.installation, (state, props) => props.component],
+  (installation, component) => installation[component.code],
 );
 
-export const getECRComponentUninstallStatus = (state, props) => (
-  get(state, `componentRepositoryComponents.uninstallation[${props.component.code}]`, '')
+export const getECRComponentUninstallStatus = createSelector(
+  [state => state.componentRepositoryComponents.uninstallation, (state, props) => props.component],
```
jeffgo10, line 68: looks like `(state, props) => props.component` is recurring on two selectors. would it be possible to put it in a separate function?

jeffgo10 (changes requested):
just one question

src/state/component-repository/components/selectors.js
ichalagashvili: Personally I prefer as it is, since its just getting single property from the props. Its more explicit like this since we just have a very simple access..

jeffgo10 (approved):
okay. approved

PR #1233, opened 2022-01-24
formik/ENG-2824 Re-implement user restrictions to use formik instead of redux-form

raxkaynan:


test/ui/users/restrictions/RestrictionsFormContainer.test.js
```diff
@@ -1,38 +1,77 @@
...
+jest.mock('state/user-settings/actions', () => ({
+  fetchUserSettings: jest.fn(() => ({ type: 'fetchUserSettings_test' })),
+  updateUserSettings: jest.fn(payload => ({ type: 'updateUserSettings_test', payload })),
 }));
 
+jest.mock('ui/users/restrictions/RestrictionsForm', () => jest.fn(mockProps => (
+  <div>
+    <button onClick={mockProps.onMount}>onMount</button>
+    <button onClick={() => mockProps.onSubmit({})}>onSubmit</button>
+  </div>
```
ichalagashvili, line 23: what are the purpose of these stuff?

raxkaynan: That's one way of testing a container component without having to fully render the child component since the child component already has its own tests. Basically, it's simulating the component's events via button clicks.
I did that to test the container component as close as possible as to how it should be tested when using RTL (even though the more appropriate approach is to fully render the children and as much as possible not using mocks, but that requires significantly more effort and seems redundant since we're already testing the child component anyway).

ichalagashvili: ok makes sense then. Thanks

ichalagashvili (approved):

PR #1239, opened 2022-01-25
formik/ENG-2819: refactor Roles screen to formik

gReis89:


src/state/roles/actions.js
```diff
@@ -66,6 +66,7 @@ export const fetchRole = roleCode => dispatch =>
       response.json().then((data) => {
         if (response.ok) {
           dispatch(initialize('role', data.payload));
```
ichalagashvili, line 68: do we still need `initialize` ? since redux-form is removed I dont see any reason to call it again?

gReis89: this same action is being used on other pages as well, so I kept it to ensure I wouldn't break anything out of my scope

src/ui/roles/common/RoleForm.js
```diff
@@ -27,154 +25,198 @@ const msgs = defineMessages({
...
+  const isEdit = mode === EDIT_MODE;
+  const superuserToggled = values.permissions ? values.permissions.superuser : false;
 
-  handleToggleSuperuser(superuserToggled) {
-    const { onToggleSuperuser, permissions } = this.props;
-    onToggleSuperuser({ superuserToggled, permissions });
-  }
+  useEffect(() => {
+    onWillMount(roleCode);
+  }, [roleCode, onWillMount]);
```
raxkaynan: This is just trivial but I think the `onWillMount` naming seems to refer more closely to the deprecated lifecycle method. Perhaps a better one would be `onDidMount` or simply just `onMount`?

raxkaynan (changes requested):
Just a minor thing.

jeffgo10 (approved):

PR #1240, opened 2022-01-27
formik/ENG-3124 modify error restructurer to accept dot notation

jeffgo10:
+ nested field code structure reverted in respect to object structure

### DIFF
--- a/src/helpers/formikUtils.js
+++ b/src/helpers/formikUtils.js
@@ -1,10 +1,11 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { FieldArray } from 'formik';
+import { get } from 'lodash';
 
 export const getTouchErrorByField = (fieldName, { touched, errors }) => ({
-  touched: touched[fieldName],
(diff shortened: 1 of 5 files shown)


src/helpers/formikUtils.js
```diff
@@ -1,10 +1,11 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { FieldArray } from 'formik';
+import { get } from 'lodash';
 
 export const getTouchErrorByField = (fieldName, { touched, errors }) => ({
-  touched: touched[fieldName],
-  error: errors[fieldName],
+  touched: get(touched, fieldName, ''),
+  error: get(errors, fieldName, ''),
```
ichalagashvili, line 8: will not it fail if `fieldName` contains dot inside?

jeffgo10: it won't. that's why I use the lodash's `get` for that specific reason - reading dot notation

ichalagashvili (approved):

PR #1266, opened 2022-03-17
ENG-3437 Fixed admin-console redirect URL

zonia3000:
Correct URL for creating a content is `do/jacms/Content/createNew.action?contentTypeCode=${contentType.typeCode}`.
When the Java app is called at this URL it sets the context and then redirects to `do/jacms/Content/entryContent.action?contentOnSessionMarker=${contentType.typeCode}_newContent`.


jeffgo10 (approved):
ah i see. thanks @zonia3000

PR #1271, opened 2022-03-21
ENG-3465 add disabled actions due to groups

ichalagashvili:
@gReis89 @raxkaynan @jeffgo10 please note that I have changed line 156 in `src/state/groups/actions.js` file. Since, `setGroups` is used for setting `myGroups`, where `groupEntries` are all available groups. The reason why it was working correctly is that we were pushing the user to group list (after creating a new group), so we are re-fetching all group entries again and it sets them correctly, so in fact it did not really matter what we left on line 156, even if we removed it it would have worked. Please give me your thought if you think otherwise


jeffgo10 (approved):
I understand what you meant. so for me, looks good 👍

raxkaynan (approved):

PR #1272, opened 2022-03-22
ENG-3463 Fix ecr not allowing installation of same bundle version

raxkaynan:


ichalagashvili (approved):
checked out locally and tested. Works well!

user-54dbb57a (approved):

PR #1282, opened 2022-04-04
ENG-3381: preventing filebrowser route to refresh path after canceling an action

gReis89:
using react-router `state` to prevent `/file-browser` route to reset path info if users click on cancel buttons

### DIFF
--- a/src/state/file-browser/actions.js
+++ b/src/state/file-browser/actions.js
@@ -71,7 +71,7 @@ export const fetchFile = (filename, extensions = ['.txt']) => (dispatch, getStat
     }
   });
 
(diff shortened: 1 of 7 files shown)


raxkaynan:
Please check [this PR](https://github.com/entando/app-builder/pull/1283). It might be solving the same thing.

EDIT: ~~I also found another issue related to my work and this one, which I'm going to fix. Is it fine with you if I address this in the changes I will make, so as to prevent conflicts? What do you think?~~

EDIT 2: Already updated [my PR](https://github.com/entando/app-builder/pull/1283) to include those changes. Please check and test to see if it also fixes the bug this PR addresses.

gReis89:
closing in favor of https://github.com/entando/app-builder/pull/1283

PR #1283, opened 2022-04-04
ENG-3376 Fix file browser breadcrumbs not working properly

raxkaynan:
Prevent action dispatch on mount of file browser page when navigating from a sub-route (i.e. via breadcrumb) since it redirects it to the root folder.

### DIFF
--- a/src/state/file-browser/actions.js
+++ b/src/state/file-browser/actions.js
@@ -85,7 +85,7 @@ export const fetchFileList = (protectedFolder = '', path = '') => dispatch =>
     getFileBrowser(`?${queryString.join('&')}`).then((response) => {
       response.json().then((json) => {
         if (response.ok) {
-          history.push(ROUTE_FILE_BROWSER);
+          history.push(ROUTE_FILE_BROWSER, { from: history.location.pathname });
           dispatch(setFileList(json.payload));
           dispatch(setPathInfo(json.metaData));
         } else {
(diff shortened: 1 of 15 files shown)


ichalagashvili (approved):
I do not fully understand what happens inside `onWillMount` but seems like it fixes the problem.

raxkaynan:
> I do not fully understand what happens inside `onWillMount` but seems like it fixes the problem.

It gets the previous location/route, then checks if it matches a sub-route of `file-browser`. If it does, then `onWillMount()` should not dispatch `fetchFileList()` since it overrides the first `fetchFileList()` dispatched from clicking a certain button (such as a breadcrumb link), which causes the file browser list to show the root directory.

chalagashvili:
@raxkaynan but you check `prevLoc === ROUTE_FILE_BROWSER ` and then dispatch `fetchFileList` , should it not be `prevLoc !== ROUTE_FILE_BROWSER` ?

raxkaynan:
> @raxkaynan but you check `prevLoc === ROUTE_FILE_BROWSER ` and then dispatch `fetchFileList` , should it not be `prevLoc !== ROUTE_FILE_BROWSER` ?

Come to think of it, it seems unnecessary to have the check, because if the previous route is the same as the current one, then `onWillMount` wouldn't be called anyway, right? I should probably remove that, but of course I need to test it first :)

gReis89 (approved):

PR #1288, opened 2022-04-12
ENG-3424 better error

ichalagashvili:
Link: https://entando.myjetbrains.com/youtrack/issue/ENG-3424
Context: If no error message provided from BE regarding deploying and undeploying bundles, show a default one


gReis89 (approved):
LGTM

PR #1290, opened 2022-04-13
ENG-3440: preventing welcome wizard to show up during same session

gReis89:
using sessionStorage to keep information between pages.

### DIFF
--- a/src/ui/app-tour/AppTour.js
+++ b/src/ui/app-tour/AppTour.js
@@ -423,9 +423,12 @@ class AppTour extends React.Component {
 
   render() {
     const {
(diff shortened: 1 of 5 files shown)


ichalagashvili:
@gReis89 what if the user clicks "Start Welcome Wizard" from top right button in the top menu? (next to license/about us). When this happens we should make session variable false, otherwise user wont be able to start wizard, because `isDismissed` will always be true

ichalagashvili (changes requested):
I think it lacks handling when user can manually trigger the wizard

gReis89:
I didn't know it was an option, good catch!

ichalagashvili (approved):

PR #1301, opened 2022-05-03
ENG-3613 implement mfe config list fetching from mocked API

ichalagashvili:


src/api/mfe.js
```diff
@@ -0,0 +1,32 @@
...
+);
+
+export const updateMfeConfig = (mfeId, version) => (
+  makeMockRequest({
+    uri: `/api/ecr/update-config/${mfeId}/${version}`,
+    method: METHODS.GET,
+    mockResponse: UPDATE_MFE_OK,
+    useAuthentication: true,
+  })
+);
```
Dv92: We are going to have only the GET API, we don't need the `update` and the `add` from the API because in our perspective the add and the update are just a store thing. The API will don't expose `add` and `update` endpoint for now

src/state/mfe/selectors.js
```diff
@@ -0,0 +1,12 @@
...
+
+
+const getMfeState = state => state.mfe;
+
+
+// eslint-disable-next-line
+export const getMfeConfigList = createSelector(
+  getMfeState,
+  mfe => mfe.mfeList,
+);
```
Dv92: can you add other two selectors as `getMfeTargetContent` and `getMfeTargetPrimaryMenu` that retrieve mfe from the config for the target value? the map it looks like this:

- getMfeContent -> target "content"
- getMfePrimaryMenu -> target "primary-menu"

This will be useful when we are going to consume this inside the layout to differentiate the lists and target the correct rendering section.

Dv92 (dismissed):

Dv92 (approved):

PR #1300, opened 2022-05-02
ENG-3584 Removed useCredentials option from getMyGroupPermissions

zonia3000:

PR #1304, opened 2022-05-09
ENG-3652 new layout for pages

ichalagashvili:
- Remove unused `useCredentials` property
- Implement new wrapper layout for pages
- Remove unused css
- Amend `InternalPage` to be a simple div

### DIFF
--- a/src/api/permissions.js
+++ b/src/api/permissions.js
@@ -20,7 +20,6 @@ export const getMyGroupPermissions = () => (
     method: METHODS.GET,
     mockResponse: MY_PERMISSIONS_PAYLOAD_OK,
     useAuthentication: true,
-    useCredentials: true,
   })
 );
 
--- a/src/ui/app/App.js
+++ b/src/ui/app/App.js
@@ -175,7 +175,7 @@ import PageNotFoundContainer from 'ui/app/PageNotFoundContainer';
 import CloneWidgetPage from 'ui/widgets/clone/CloneWidgetPage';
 import EmailConfigPage from 'ui/email-config/EmailConfigPage';
 
-import InternalPage from 'ui/internal-page/InternalPage';
+import PageLayout from 'ui/app/PageLayout';
 import RowSpinner from 'ui/pages/common/RowSpinner';
 import entandoApps from 'entando-apps';
 import AboutPage from 'ui/about/AboutPage';
@@ -191,7 +191,7 @@ const appsRoutes = entandoApps.reduce((routes, app) => (
         key={AppRoute.path}
         path={AppRoute.path}
         // eslint-disable-next-line react/jsx-pascal-case
-        render={() => <InternalPage><AppRoute.component /></InternalPage>}
+        render={() => <PageLayout><AppRoute.component /></PageLayout>}
       />
     )),
   ]
@@ -208,7 +208,7 @@ const getRouteComponent = () => {
           <LoginPage>
             <LoginFormContainer />
           </LoginPage>
-      )}
+        )}
       />
       <Route path={ROUTE_DASHBOARD} component={DashboardPage} />
       {/* page */}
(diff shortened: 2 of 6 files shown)


src/ui/app/PageLayout.js
```diff
@@ -0,0 +1,36 @@
...
+import ActivityStreamContainer from 'ui/activity-stream/ActivityStreamContainer';
+import NotificationListContainer from 'ui/activity-stream/NotificationListContainer';
+import VerticalMenuContainer from 'ui/internal-page/VerticalMenuContainer';
+import AboutInfoModal from 'ui/about/AboutInfoModal';
+import LicenseInfoModal from 'ui/license/LicenseInfoModal';
+
+const PageLayout = ({ className, children }) => (
+  <div
+    data-testid="page-layout"
+    className={[className, 'layout-pf-fixed'].join(' ').trim()}
```
Dv92: I think we don't need prop className, we can remove it

src/ui/app/PageLayout.js
```diff
@@ -0,0 +1,36 @@
...
+);
+
+PageLayout.propTypes = {
+  children: PropTypes.node,
+  className: PropTypes.string,
+};
+
+PageLayout.defaultProps = {
+  children: null,
+  className: '',
```
Dv92: can remove also this defaultProps

src/ui/app/App.js
```diff
@@ -191,7 +191,7 @@ const appsRoutes = entandoApps.reduce((routes, app) => (
         key={AppRoute.path}
         path={AppRoute.path}
         // eslint-disable-next-line react/jsx-pascal-case
-        render={() => <InternalPage><AppRoute.component /></InternalPage>}
+        render={() => <PageLayout><AppRoute.component /></PageLayout>}
```
Dv92: here you don't need to wrap AppRoute.component with PageLayout because is already inherited inside the component App.

src/api/permissions.js
```diff
@@ -20,7 +20,6 @@ export const getMyGroupPermissions = () => (
     method: METHODS.GET,
     mockResponse: MY_PERMISSIONS_PAYLOAD_OK,
     useAuthentication: true,
```
Dv92, line 22: why do you need to delete useCredentials? It's related to this task? are you sure that this doesn't take unexpected issues?

Dv92:
@ichalagashvili the test for PageLayout.js is also missing, you should add

src/api/permissions.js
Dv92: I see there is a PR of @zonia3000 about that https://github.com/entando/app-builder/pull/1300

src/ui/app/PageLayout.js
chalagashvili: done ✅

src/ui/app/PageLayout.js
chalagashvili: done ✅

src/ui/app/App.js
chalagashvili: done ✅

src/api/permissions.js
ichalagashvili: no any other API call uses `useCredentials: true`. Removing it does not interfere with anything, its simply not necessary. On the other hand, when using local app-builder with remote de-app instance, that line causes CORS errors, so it has to be definitely removed.

Dv92: ok, talk with @zonia3000 and agree on her PR. She has made the same fix.

zonia3000: Hi, it is ok for me to remove `useCredentials` on this PR. I can close (without merging) my PR when this one will be merged.

Dv92 (approved):

src/ui/app/App.js
```diff
@@ -175,12 +175,13 @@ import PageNotFoundContainer from 'ui/app/PageNotFoundContainer';
...
+import PageLayout from 'ui/app/PageLayout';
 import RowSpinner from 'ui/pages/common/RowSpinner';
 import entandoApps from 'entando-apps';
 import AboutPage from 'ui/about/AboutPage';
 import LicensePage from 'ui/license/LicensePage';
 import getRuntimeEnv from 'helpers/getRuntimeEnv';
+import InternalPage from 'ui/internal-page/InternalPage';
 
 const appsRoutes = entandoApps.reduce((routes, app) => (
   [
```
gReis89, line 187: on line 194, we are still using InternalPage as a wrapper on a render method, shouldn't it be replaced by the PageLayout as well?

gReis89: nevermind, looks like it is already being wrapped by the PageLayout 👍

gReis89 (approved):

PR #1306, opened 2022-05-10
ENG-3589 ENG-3680 sort versions descending

ichalagashvili:


src/ui/component-repository/components/item/install-controls/InstallButton.js
```diff
@@ -18,6 +18,28 @@ const jobProgressStatuses = [
...
+    if (a2 !== b2) {
+      return a2 > b2 ? 1 : -1;
+    }
+  }
+
+  // 4. We hit this if the all checked versions so far are equal
+  //
+  return b1.length - a1.length;
+};
+
```
Dv92: I think you should move this function inside the helper directory (inside helper/versions.js I think can be a good place). You should also add the tests for this function.

Dv92 (approved):

PR #1313, opened 2022-05-17
ENG-3584 remove unused useCredentials line

ichalagashvili:


Dv92:
After a talk with @cecchisandrone and @zonia3000 we agree that the error of CORS that this useCredentials seems to resolve is a separate backend issue that need investigation. The backend part of ENG-3584 is **not on develop** so we can't merge safety this fix because it's not related to this ticket. @ichalagashvili please open a ticket where you describe the CORS error with remote instances of de-app so the backend team can address the issue and find a solution.

PR #1316, opened 2022-05-19
ENG-3697 change contract gui

ichalagashvili:


test/test/mocks/mfe.js
```diff
@@ -1,65 +1,78 @@
...
+        name: 'ms1',
+        baseUrl: 'http://localhost:8080',
+      },
+      {
+        name: 'ms2',
+        baseUrl: 'http://localhost:8090',
       },
-    },
-    activeRoutes: ['/example-mfe'],
+    ],
```
Dv92: for this one, we need to keep the data related to `example-mfe` (it's the local mock)  but with the new structure

Dv92 (approved):

PR #1318, opened 2022-05-24
ENG-3754  update Web Components presentation and navigation for the sprint demo 

Dv92:


src/hooks/useMfe.js
```diff
@@ -4,11 +4,8 @@ import { getMfeById } from 'state/mfe/selectors';
 
 
 // try to find the element by id
-const isLoaded = (id) => {
-  const script = document.getElementById(`script-${id}`);
-  const style = document.getElementById(`style-${id}`);
-  return script && style;
-};
+const isJSLoaded = id => document.getElementById(`script-${id}`);
+const isCSSLoaded = id => document.getElementById(`style-${id}`);
```
Dv92, line 8: the CSS asset can be optional. Some MFEs don't export an external CSS because they inject it directly into the shadow dom. For this reason, we need to decouple the control regarding the check for the style loading. This prevents a bug of infinite loop when an MFE without external CSS is loaded.

src/hooks/useMfe.js
```diff
@@ -4,11 +4,8 @@ import { getMfeById } from 'state/mfe/selectors';
 
 
 // try to find the element by id
-const isLoaded = (id) => {
-  const script = document.getElementById(`script-${id}`);
-  const style = document.getElementById(`style-${id}`);
-  return script && style;
-};
+const isJSLoaded = id => document.getElementById(`script-${id}`);
```
ichalagashvili, line 7: I wonder maybe we need some kind of `retry with timeout` or not. Is it possible that JS file is not loaded during this check?

ichalagashvili (approved):
LGTM, Just a one question

src/hooks/useMfe.js
Dv92: In this case, we only check if the DOM element of the script is on the page. We didn't have this problem because the script's loading is not related to its DOM injection. The check here is only for preventing from inserting multiple times the same scripts.

gReis89 (approved):
LGTM!

PR #1323, opened 2022-05-26
ENG-3772: fixing Entando default header

gReis89:
For some weird reason, the component is not being rendered on line 235 as a React component, but it does work when turning it into a function.

### DIFF
--- a/src/ui/internal-page/VerticalMenuContainer.js
+++ b/src/ui/internal-page/VerticalMenuContainer.js
@@ -164,7 +164,7 @@ const renderComponentRepositoryMenuItem = (history, intl) => (
     title={intl.formatMessage({ id: 'componentRepository.menuButton.title' })}
   />) : '');
 
-const Header = ({ onStartTutorial }) => (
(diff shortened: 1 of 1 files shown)


ichalagashvili (approved):

Dv92 (approved):
@gReis89 I've checked the code of `VerticalNav`; the `Header` component it's not rendered because the parent (`VerticalNav`) expect `MasterHead` as direct child. When we use Header as component the tree is like this:

VerticalNav -> Header -> MasterHead

Your solution work because it change the relationship as VerticalNav expected:

VerticalNav -> MasterHead

Dv92:
> @gReis89 I've checked the code of `VerticalNav`, `Header` component it's not rendered because the parent (`VerticalNav`) expect `MasterHead` as direct child. When we use Header as component the tree is like this:
> 
> VerticalNav -> Header -> MasterHead
> 
> Your solution work because it change the relationship as VerticalNav expected:
> 
> VerticalNav -> MasterHead

@ichalagashvili FYI

PR #1325, opened 2022-05-27
ENG-3663 fix breaking edit PageTemplate page

ichalagashvili:
- Break was happening because of `codemirror` library. If we provide object it does not work, we need to give stringified object always.


Dv92 (dismissed):

gReis89 (approved):
looks good

Dv92 (approved):

PR #1329, opened 2022-06-09
ENG-3792 change mfe routing path assembling

ichalagashvili:


src/helpers/urlUtils.js
```diff
@@ -1,3 +1,14 @@
 import { get } from 'lodash';
-// eslint-disable-next-line import/prefer-default-export
+
 export const adminConsoleUrl = url => `${get(process.env, 'DOMAIN', '')}/${url}`;
+
+export const generateMfeRoutes = mfe => mfe.reduce((acc, curr) => {
+  if (curr.paths) {
+    const routes = [];
+    curr.paths.forEach((route) => {
+      routes.push({ route: `/${curr.pbcName}/${curr.bundleName}${route}`, id: curr.id || `id-${curr.pbcName}-${curr.bundleName} - ${route}` });
```
Dv92: sometimes `route` can start with `/` and other times could be without, this depends on how the user have written the path in the bundle descriptor. So we must prevent issues in the interpolation, for this reason I suggest you to clean the route and create the url as a `join` in this way:
```javascript
const path = route.split("/").filter(Boolean).join("/")
const route = [curr.pbcName, curr.bundleName, path].join("/")

routes.push({ route: `/${route}`, id: curr.id ||`id-${curr.pbcName}-${curr.bundleName}-${route}` });
```
you can do the same thing for the `id` but with "-", also remove the spaces inside the `id`

ichalagashvili: done ✅

Dv92 (dismissed):

Dv92:
@ichalagashvili please, can you see the coverage issue?

src/ui/app/App.js
```diff
@@ -205,17 +206,12 @@ const RouteComponent = () => {
...
-        routes.push({ route, id: curr.id });
-      });
-      return [...acc, ...routes];
-    }
-    return acc;
+  const mfeRoutes = React.useMemo(
+    () => generateMfeRoutes(contentMfe),
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, []), [JSON.stringify(contentMfe)]);
+    [],
```
gReis89: should it be removed? looks like you are passing 3 parameters to `useMemo` instead of 2.

src/ui/app/App.js
```diff
@@ -205,17 +206,10 @@ const RouteComponent = () => {
...
-        routes.push({ route, id: curr.id });
-      });
-      return [...acc, ...routes];
-    }
-    return acc;
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, []), [JSON.stringify(contentMfe)]);
+  const mfeRoutes = React.useMemo(
+    () => generateMfeRoutes(contentMfe),
+    [contentMfe],
```
gReis89: this is an array of object, I think it should be stringified in order to correctly memoise it, what do u think @Dv92 ?

gReis89 (dismissed):
looks good!

gReis89 (dismissed):
LGTM

gReis89 (approved):

PR #1332, opened 2022-06-13
ENG-3663: fixing page designer

gReis89:
After merging the [PR](https://github.com/entando/app-builder/pull/1325) with the fix for [ENG-3663](https://entando.myjetbrains.com/youtrack/issue/ENG-3663), the Page Designer started to crash, because it needs the PageTempalte configs to be an object in the redux store. I rolled back some of the changes and fixed the issue on a different way to keep the redux state untouched.


ichalagashvili:
@gReis89 @Dv92 this PR should be closed since the fix for this is already in there. It was merged 4 days ago, https://github.com/entando/app-builder/pull/1328 so probably you did not have latest develop branch. It was also tested by QE and moved to stage.

gReis89:
ops, that's true, I'm going to close it.

PR #1337, opened 2022-06-15
ENG-3819 modified shell globals

ichalagashvili:


src/ui/app/MfeContainer.js
```diff
@@ -6,23 +6,37 @@ import PropTypes from 'prop-types';
...
   const locale = useSelector(getLocale);
   const permissions = useSelector(getLoggedUserPermissions);
+  const systemReport = useSelector(getSystemReport);
 
   useEffect(() => {
     if (!window.appBuilderRouter) {
       window.appBuilderRouter = history;
     }
-  }, [history]);
+    if (!window.entando || !window.entando.globals) {
```
gReis89: this if condition will prevent the `window.entando` object to get updated(for example, if locale changes), is that what we want?

Dv92: no, globals must be updated if some values changed.

gReis89 (approved):
LGTM

PR #1339, opened 2022-06-21
ENG-3845 rename router for mfe

ichalagashvili:


src/ui/app/MfeContainer.js
```diff
@@ -16,10 +16,11 @@ const MfeContainer = ({ id, history }) => {
   const systemReport = useSelector(getSystemReport);
 
   useEffect(() => {
-    if (!window.appBuilderRouter) {
-      window.appBuilderRouter = history;
-    }
     const entandoWindow = window.entando || {};
+    if (!entandoWindow.router) {
+      window.entando = { ...entandoWindow, router: history };
+      entandoWindow.router = history;
```
gReis89, line 21: having to set the `window.entando` and the `entandoWindow` is a bit weird.
Maybe if we reorganize the code, we can create the `entandoWindow` as needed and then set it to the `window.entando` at the end just once

ichalagashvili: agree, done ✅

gReis89 (approved):

PR #1344, opened 2022-06-29
ENG-3851: Entando installation screen

gReis89:


Dv92:
@gReis89 this PR must not be merged until QE ends the cypress test update. See YT task for clarifications

sass/pages/common/NewSpinner.scss
```diff
@@ -0,0 +1,21 @@
...
+  }
+}
+
+@keyframes spin {
+  from {
+    transform: rotate(0deg);
+  }to {
+    transform: rotate(360deg);
+  }
+}
```
ichalagashvili: sorry to be that guy but new line please..

src/ui/app-tour/AppTour.js
```diff
@@ -39,8 +39,6 @@ class AppTour extends React.Component {
   }
 
   componentDidMount() {
-    const { onDidMount } = this.props;
-    onDidMount(this.props);
```
ichalagashvili: this should not have been inside this commit..

ichalagashvili (changes requested):
Just small things

src/app-init/MfeDownloadManager.js
```diff
@@ -3,24 +3,43 @@ import { useDispatch, useSelector } from 'react-redux';
...
+        dispatch(fetchInitialMfeConfigList()).then(() => {
+          setLoading(false);
+          setIsPolling(false);
+        }).catch(() => {
+          // if fails, start polling until it succeed
+          setLoading(false);
+          setIsPolling(true);
+          setTimeout(fetchConfig, 10000);
+        });
+      };
```
Dv92: you need to [clear](https://developer.mozilla.org/en-US/docs/Web/API/clearTimeout) `setTimeout` after the execution, this can cause bugs that are uneasy to find out.

src/app-init/StartupWaitScreen.js
```diff
@@ -0,0 +1,12 @@
...
+
+const StartupWaitScreen = () => (
+  <div className="StartupWaitScreen">
+    <NewSpinner loading />
+    <h2>Entando is installing</h2>
+    <p>Wait a few more minutes for the installation to complete</p>
+  </div>
+);
+
+export default StartupWaitScreen;
```
Dv92: can you move this component inside the `ui/app` folder?

src/state/mfe/actions.js
```diff
@@ -35,3 +35,19 @@ export const fetchMfeConfigList = (page = { page: 1, pageSize: 0 }, params = '')
...
+        response.json().then((json) => {
+          if (response.ok) {
+            dispatch(setMfeConfigList(json.payload));
+            resolve(response);
+          } else {
+            reject();
+          }
+        }).catch(() => reject());
+      }).catch(() => reject());
+  });
```
Dv92: instead of duplicating the action, why don't add a boolean param `withToastNotification`  to `fetchMfeConfigList` to disable the toasts notification? the logic is the same.

src/app-init/MfeDownloadManager.js
gReis89: humm I never heard about clearing a setTimeout... I did small research and looks like it is not really necessary:
https://stackoverflow.com/questions/7391567/when-using-settimeout-do-you-have-to-cleartimeout#:~:text=You%20don't%20actually%20need,because%20setInterval%20usually%20runs%20indefinitely.

gReis89: since we are not storing the timeout instance, I guess the clearing doesn't need to happen. If you look at the docs example, it is storing the setTimeout instance on an object attribute, which is different from what we are doing here.

src/state/mfe/actions.js
gReis89: the implementation is slightly different as well.. I also change the way we resolve/reject the promises, I guess adding if/else statement would turn the method very difficult to read.

this is how it would look like:

```js
export const fetchMfeConfigList = (page = { page: 1, pageSize: 0 }, params = '', withToastNotification) => dispatch =>
  new Promise((resolve, reject) => {
    getMfeConfigList(page, params)
      .then((response) => {
        response.json().then((json) => {
          if (response.ok) {
            dispatch(setMfeConfigList(json.payload));
            if (!withToastNotification) {
              resolve(response);
            }
          } else if (withToastNotification) {
            json.errors.forEach(err => dispatch(addToast(err.message, TOAST_ERROR)));
          } else {
            reject();
          }
        }).catch((error) => {
          if (withToastNotification) {
            dispatch(addToast(error.message, TOAST_ERROR));
          } else {
            reject();
          }
        });
      }).catch((error) => {
        if (withToastNotification) {
          dispatch(addToast(error.message, TOAST_ERROR));
        } else {
          reject();
        }
      });
  });
```

src/app-init/MfeDownloadManager.js
Dv92: my concern is that it is a polling case we will fulfill [the event loop](https://developer.mozilla.org/en-US/docs/Web/JavaScript/EventLoop) with different `setTimeout` that are riding at different times. If you see [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/EventLoop#adding_messages) the delay is the _minimum_ time and depends on how many messages are in the queue, this can cause a performance regression as happen for the loops. Maybe it's premature optimization, I know, so for now you can omit the modification but keep in mind this possibility.

src/state/mfe/actions.js
Dv92: If you want to improve the readability you could use a functional approach, for example:

```javascript
function notify({message, enabled}, cb) {
    if (enabled) return dispatch(addToast(message, TOAST_ERROR));

    return cb ? cb() : null
}

export const fetchMfeConfigList = (page = {page: 1, pageSize: 0}, params = '', withToastNotification) => dispatch =>
    new Promise((resolve, reject) => {
        getMfeConfigList(page, params)
            .then((response) => {
                response.json().then((json) => {
                    if (response.ok) {
                        dispatch(setMfeConfigList(json.payload));
                        resolve(response);
                    } else if (withToastNotification) {
                        json.errors.forEach(err => notify({ message: err.message, enabled: true }));
                    }

                    reject()
                }).catch((error) => {
                   notify({ message: error.message, enabled: withToastNotification }, reject)
                });
            }).catch((error) => {
                notify({ message: error.message, enabled: withToastNotification }, reject)
            });
    });
```

Regarding the `resolve` and `reject`, I think you could normalize also. From what I see `fetchMfeConfigList` it's not used with handlers on `then` and `catch` so we could set the behavior without problems.

Dv92 (approved):

PR #1346, opened 2022-07-01
ENG-3908 infinite loop fix

ichalagashvili:


gReis89 (dismissed):

src/hooks/useMfe.js
```diff
@@ -63,7 +63,9 @@ const injectAssetToDom = ({ id, assets = [] }, remove, setError) => {
 
 const useMfe = (mfeId) => {
   // if mfeId is already an object, return it, otherwise select it from store
-  const mfe = useSelector(state => (typeof mfeId === 'object' ? mfeId : getMfeById(state, mfeId)));
+  const mfe = useSelector(state => getMfeById(state, mfeId));
+
+  const memoMfe = useMemo(() => (typeof mfeId === 'object' ? mfeId : mfe), [mfe, mfeId]);
```
Dv92: I expect that `mfeId` is always a primitive value (string in this case), why this `typeof object`? I see that is related to another PR merged these days. If we need to pass a `mfe` object from outside is better to add another param (for example `initialMfe`). Can you do this modification?

Dv92 (commented):
can you checkout the code unrelated to the fix?

src/hooks/useMfe.js
ichalagashvili: refactor applied ✅

Dv92 (approved):

PR #1350, opened 2022-07-07
ENG-3621 change route path

ichalagashvili:


gReis89 (dismissed):

src/api/mfe.js
```diff
@@ -5,7 +5,7 @@ import { LIST_MFE_RESPONSE_OK } from 'test/mocks/mfe';
 export const getMfeConfigList = (page = { page: 1, pageSize: 0 }, params = '') => (
   makeMockRequest(
     {
-      uri: `api/ecr/config-list${params}`,
+      uri: `/bundles/all/widgets?widgetType="app-builder"${params}`,
       method: METHODS.GET,
       mockResponse: LIST_MFE_RESPONSE_OK,
       useAuthentication: true,
```
Dv92: can you update the endpoint as below? we also need to change `makeMockRequest` to `makeRequest` and delete the page object. I'll update the task 

```js
export const getMfeConfigList = (params = ' ') => (
  makeRequest({
    uri: `/bundles/all/widgets?filters[0].value=app-builder&filters[0].attribute=widgetType&filters[0].operator=eq${params}`,
    domain: '/digital-exchange',
    method: METHODS.GET,
    mockResponse: LIST_MFE_RESPONSE_OK,
    useAuthentication: true,
  })
);
```

ichalagashvili: done ✅

Dv92 (approved):

PR #1355, opened 2022-07-14
ENG-3952: Pagination not correctly updated in roles, groups, profile types, fragments

riccardo-fadda:


src/state/roles/actions.js
```diff
@@ -142,7 +143,7 @@ export const fetchUserRefs = (roleCode, page = { page: 1, pageSize: 10 }, params
       response.json().then((data) => {
         if (response.ok) {
           dispatch(setUserRefs(data.payload));
-          dispatch(setPage(data.metaData));
+          if (data.payload.length > 0) dispatch(setPage(data.metaData));
```
ichalagashvili, line 146: so if we delete last user ref, then pagination will still say 1, instead of 0. So, can you please tell why we need this `if` checker?

ichalagashvili (commented):
Looks good to me. Just one question

src/state/roles/actions.js
riccardo-fadda: Without it, when selecting deletion of a role, if the role has no references the User Reference table won't be shown, and the pagination of the role tables will  instead reflect the results of getUserReferences (1-0 of 0)

ichalagashvili (approved):

Dv92 (approved):

PR #1401, opened 2022-09-05
ENG-4068 Add MfeDownloadManager tests

raxkaynan:


ichalagashvili (approved):

test/app-init/MfeDownloadManager.test.js
```diff
@@ -0,0 +1,38 @@
...
+    expect(screen.getByTestId('spinner')).toBeInTheDocument();
+  });
+
+  it('shows startup wait screen when polling', async () => {
+    fetchMfeConfigList.mockImplementationOnce(() => () => new Promise(resolve => resolve({})));
+
+    renderWithState(<MfeDownloadManager>test</MfeDownloadManager>);
+
+    expect(await screen.findByTestId('startup-wait-screen')).toBeInTheDocument();
+  });
```
Dv92, line 28: I would add another test to check the same behavior when the result is `null` or `undefined` (it should happen when the API has a status code of 500).

raxkaynan: It seems that response won't ever be `null` or `undefined` since it comes from `fetch`. Also, when a `5xx` error occurs, the Promise gets rejected. So I just added a test for that scenario.

Dv92 (approved):

PR #1421, opened 2022-11-17
ENG-4259 recover enable save button code

ichalagashvili:


src/ui/widgets/config/WidgetConfigMicrofrontend.js
```diff
@@ -25,6 +25,21 @@ const WidgetConfigMicrofrontend = ({
   const { assetLoading, mfe, hasError } = useMfe({ initialMfe: memoMfeObject });
   const customElement = get(mfe, 'customElement');
 
+  // eslint-disable-next-line no-unused-vars
+  const [enableSubmit, setEnableSubmit] = useState(true);
```
raxkaynan, line 28: `[, setEnableSubmit]` can be used here so the eslint comment can be removed.

Also, what's the purpose of using `setEnableSubmit()` if its corresponding state value isn't used?

raxkaynan (changes requested):

src/ui/widgets/config/WidgetConfigMicrofrontend.js
Dv92: @chalagashvili from what I see in the history, it's missing the `enableSubmit` binding to the save button. It's never been added 😅. So we need that property @raxkaynan but must be hook to the save button for disable when it is false.

Dv92 (changes requested):
need the modification in the comment above

Dv92:
moved to draft until 7.1.2 release

Dv92 (approved):

PR #1422, opened 2022-11-21
ENG-4334 implement correct semantic versions checker

ichalagashvili:
- Fix small css issue

### DIFF
--- a/sass/component-repository/components/ComponentList.scss
+++ b/sass/component-repository/components/ComponentList.scss
@@ -45,7 +45,7 @@
 
   &__buttons-container {
     display: flex;
-    max-height: 24px !important; // sass-lint:disable-line no-important
+    max-height: 24px;
 
     & > div {
       margin-right: 10px;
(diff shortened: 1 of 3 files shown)


raxkaynan (dismissed):

sass/component-repository/components/ComponentList.scss
```diff
@@ -45,6 +45,7 @@
 
   &__buttons-container {
     display: flex;
+    max-height: 24px !important; // sass-lint:disable-line no-important
```
Dv92: why do you need this important? are you sure it is necessary?

Dv92:
convert to draft until 7.1.2 release

sass/component-repository/components/ComponentList.scss
ichalagashvili: You are right, no need for important here. Checked without it and all good.

Dv92 (approved):

PR #1444, opened 2023-01-05
ENG-4459 fix filter for widgets list

ichalagashvili:


src/ui/pages/config/WidgetGroupings.js
```diff
@@ -14,7 +14,8 @@ const msgs = defineMessages({
 });
 
 const WidgetGroupings = ({
-  intl, groupedWidgets, widgetGroupingList, filterWidget, locale, searchFilter,
+  // eslint-disable-next-line no-unused-vars
```
raxkaynan: It seems there's no unused variable here.

raxkaynan (changes requested):

src/ui/pages/config/WidgetGroupings.js
```diff
@@ -35,13 +36,16 @@ const WidgetGroupings = ({
     setOpenGroupings(newOpenGroupings);
   }, [searchFilter, widgetGroupingList]);
 
+  useEffect(() => () => {
+    clearFilter();
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
Dv92, line 41: @ichalagashvili I'm thinking aloud, the problem is when someone set the filter, move forward and then come back. I think that instead of at mount, the `clearFilter` should be called on [cleanup](https://reactjs.org/docs/hooks-effect.html#example-using-hooks-1), what do you think?

ichalagashvili: @Dv92 its called on cleanup. as you see its is `() => () => {` so, its during retuning of useEffect, so its cleanup, not during on mount

src/ui/pages/config/WidgetGroupings.js
ichalagashvili: done

src/ui/pages/config/WidgetGroupings.js
Dv92: ouch, I didn't see sorry 😅

Dv92 (dismissed):

Dv92 (approved):

raxkaynan (approved):

PR #1449, opened 2023-01-17
ENG-4362 add proper validations

ichalagashvili:


src/helpers/attrValidation.js
```diff
@@ -154,6 +154,18 @@ export const regex = text => (value) => {
...
+  return !re.test(value)
+    ? (
+      <FormattedMessage
+        id="validateForm.email"
+        values={{ regex: <b>{text}</b> }}
+        defaultMessage="Invalid email address."
+      />
+    ) : undefined;
+};
+
```
Dv92: in the project we have as dependency [`validator` ](https://www.npmjs.com/package/validator) you can use `isEmail` for a correct validation of the value

ichalagashvili: done

Dv92 (approved):

PR #1450, opened 2023-01-19
ENG-4487 hide menu and header if forbidden access

ichalagashvili:


Dv92 (approved):
LGTM, it's not the perfect solution as we discussed @ichalagashvili. We will fix the `PermissionCheck` design in the other ticket

PR #1452, opened 2023-01-26
ENG-4362 refactor validations and fix tech debt

ichalagashvili:
- Increased `minLength10` to `minLength15` for Profile Type Attribute code size, since `profilepicture` has length of 14 and its a default attribute.
- Replaced `field` renderer with existing `UserProfileField` and deleted repeated code.
- Point 2 fix also fixed the validations that are defined by users for each field for each profile type attributes.

fyi: the other PR was closed as it had incorrect solution

### DIFF
--- a/src/ui/common/attributes/AttributeInfo.js
+++ b/src/ui/common/attributes/AttributeInfo.js
@@ -9,7 +9,7 @@ import FormLabel from 'ui/common/form/FormLabel';
 import SwitchRenderer from 'ui/common/form/SwitchRenderer';
 import { MODE_EDIT, MODE_ADD } from 'state/data-types/const';
 
-const maxLength10 = maxLength(10);
+const maxLength15 = maxLength(15);
 const maxLength50 = maxLength(50);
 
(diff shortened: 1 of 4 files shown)


src/ui/common/attributes/AttributeInfo.js
```diff
@@ -9,7 +9,7 @@ import FormLabel from 'ui/common/form/FormLabel';
 import SwitchRenderer from 'ui/common/form/SwitchRenderer';
 import { MODE_EDIT, MODE_ADD } from 'state/data-types/const';
 
-const maxLength10 = maxLength(10);
+const maxLength15 = maxLength(15);
 const maxLength50 = maxLength(50);
```
Dv92, line 13: changing this value seems have a global impact to other part of the code that use `AttributeInfo`. Are you sure that this doesn't cause bug in other sections?

ichalagashvili: yes its safe

Dv92 (approved):

DMinnai (approved):

PR #1454, opened 2023-02-15
ENG-4606: EPC MFE routing issue

Dv92:
- reinforce the `key` prop on the dynamic route of EPC MFE for avoid the rendering issue


ichalagashvili (approved):

Dv92:
The problem mentioned here is a false positive. The routing issue is caused by the routing implementation inside the EPC. The fix will be done on the EPC.

PR #1461, opened 2023-03-28
ENG-4652 update code valid

DMinnai:


src/helpers/attrValidation.js
```diff
@@ -215,3 +215,12 @@ export const imageValidate = memoize(langCode => (input) => {
...
 }, (...args) => JSON.stringify(args));
+
+export const codeWithDash = value =>
+  (value && /^[0-9a-zA-Z_.-]+$/i.test(value) ? undefined : (
+    <FormattedMessage id="validateForm.codeWithDash" />
+  ));
+export const friendlyCodeWithDash = value =>
+  (value && /^[0-9a-zA-Z_.-]+$/i.test(value) ? undefined : (
+    <FormattedMessage id="validateForm.friendlyCodeWithDash" />
+  ));
```
ichalagashvili, line 226: why do we need 2 separate validators? is not just `codeWithDash` enough?

ichalagashvili (changes requested):
please fix the eslint/prettier changes.

src/helpers/attrValidation.js
DMinnai: I did include 2 different validator because intl messages included are different.

ichalagashvili (approved):

src/helpers/attrValidation.js
DMinnai: I did reopen the pull for the necessary test

PR #1473, opened 2023-04-14
ENG-4626 added removed labels on Dashboard

DMinnai:


ichalagashvili (changes requested):
please also include pt.js (portugese)

DMinnai:
> please also include pt.js (portugese)

Pt misses all the cms labels.. to i Have to add them all?

ichalagashvili (approved):

PR #1528, opened 2023-08-08
formik/ENG-4407 nav bar

ryanjpburgos:


src/ui/widgets/config/forms/NavigationBarConfigFormContainer.js
```diff
@@ -109,6 +109,12 @@ export const mapDispatchToProps = (dispatch, ownProps) => ({
       dispatch(setAppTourLastStep(16));
     }
   },
+  onRemoveExpression: (index) => {
```
ichalagashvili: `reduxForm` library should be totally removed from this file. Any functionality (e.g `arrayRemove` ) should be replaced with `formik`'s alternative, or implemented some other way if it does not exist. Main goal is to at the end of the activities to uninstall `redux-form` library once all our formik tickets are done, so thats why we need remove any reference with `redux-form`

ichalagashvili (changes requested):
just 2 functions need to be replaced from redux-form and wrapping the component with redux-form should be removed

DMinnai (approved):

src/ui/widgets/config/forms/NavigationBarConfigFormContainer.js
```diff
@@ -88,40 +83,48 @@ export const mapDispatchToProps = (dispatch, ownProps) => ({
...
+    expressions.splice(index, 1);
+    setFieldValue('expressions', expressions);
+  },
+  onSwapExpressions: (indexA, indexB, setFieldValue, values) => {
+    const { expressions } = values;
+    const temp = expressions[indexA];
+    expressions[indexA] = expressions[indexB];
+    expressions[indexB] = temp;
+    setFieldValue('expressions', expressions);
+  },
```
ichalagashvili: I feel a bit uneasy when we are directly modifying the `expressions` . Because `expressions` come from `values` , which comes from `formik`'s internal memory, so we are directly changing array inside the formik memory. It works yes, but in some moment it might cause some unexpected behaviour. So I would do like this for example:

```javascript
  onRemoveExpression: (index, setFieldValue, values) => {
    const { expressions } = values;
    const newExpressions = [...expressions];
    newExpressions.splice(index, 1);
    setFieldValue('expressions', newExpressions);
  }
```
same with the `onSwapExpressions`

ichalagashvili (changes requested):
just one thing 🙌

ichalagashvili (approved):

PR #1543, opened 2023-09-13
ENG-5080 only expand the my_homepage if the tour has started

ichalagashvili:
- if my_homepage is available then show wizard, if not do NOT show

### DIFF
--- a/src/state/pages/actions.js
+++ b/src/state/pages/actions.js
@@ -195,13 +195,7 @@ export const fetchPageInfo = wrapApiCall(SEO_ENABLED ? getPageSEO : getPage);
 export const fetchPageChildren = wrapApiCall(getPageChildren);
 
 export const fetchIfPageExists = pageCode => new Promise((resolve) => {
-  getPage(pageCode).then((response) => {
-    if (response.ok) {
-      resolve(true);
-    } else {
-      resolve(false);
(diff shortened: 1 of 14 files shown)


ryanjpburgos (dismissed):

src/state/pages/actions.js
```diff
@@ -194,6 +194,17 @@ export const fetchPage = wrapApiCall(getPage);
 export const fetchPageInfo = wrapApiCall(SEO_ENABLED ? getPageSEO : getPage);
 export const fetchPageChildren = wrapApiCall(getPageChildren);
 
+export const fetchIfPageExists = pageCode => new Promise((resolve) => {
+  getPage(pageCode).then((response) => {
+    if (response.ok) {
+      resolve(true);
+    } else {
+      resolve(false);
+    }
```
ryanjpburgos: maybe here you can resolve directly response.ok

```suggestion
    resolve(response.ok)
```

although I think that if it's not ok it'll go directly to the catch maybe

ichalagashvili: done

ichalagashvili: no it wont go, depends on error code.

ryanjpburgos (approved):

PR #1612, opened 2026-03-09
ESB-1012 - Render dynamic CMS menu items from SystemReport API response

ffalqui:
- hardcoded scheduler/workflow menu items with a dynamic loop
   - over the new array-based SystemReport payload, filtering by permissions.


MEM2677 (approved):
LGTM

